Issue #3, opened 2015-03-27
Theme: Hydrogen - Please add css bootstrap 2.X


810:
Could you add some extra css to support the default bootstrap css on joomla.

This is what missing:
- hidden-phone, visible-phone, ect
- textarea: width is wrong
- Table widths

Solution 1:
````
@media (min-width: 960px)  {
 .visible-phone {
  display: none;
 }
}
````
Solution 2:
/media/gantry5/engines/nucleus/scss/joomla/_forms.scss
````
input, textarea, .uneditable-input {
    width: 90%;
    Remove * max-width: 250px;
}
````

810:
Remove /media/gantry5/engines/nucleus/scss/joomla/_offline.scss:
````
#system-message {
    width:50%;
}
````

w00fz:
Portion of this was fixed by @kevlardev in 85ee671e22 and b30bc88fc7

810:
All is merged closing this one

PR #5, opened 2015-03-28
Fix width system message and fix center #3

810:


810:
is already merged, closing my pr

PR #128, opened 2015-04-27
Gantry 4 Preset

Omegatcu:
Added Preset for old Gantry 4 Configuration

### DIFF
--- a/engines/joomla/nucleus/layouts/_joomla_-_gantry4.yaml
+++ b/engines/joomla/nucleus/layouts/_joomla_-_gantry4.yaml
@@ -1,26 +1,26 @@
+drawer:
+  container:
+    - [position-drawer 100]
+
 top:
   container:
-    - [position-top-a 20,position-top-b 20,position-top-c 20,position-top-d 20,position-top-e 20]
+    - [position-top-a 33.3,position-top-b 33.3,position-top-c 33.3]
 
 navigation:
   container:
     - [position-header-a 30, particle-menu 70]
 
-drawer:
-  container:
-    - [position-drawer 100]
-
 showcase:
   container:
-    - [position-showcase-a 20,position-showcase-b 20,position-showcase-c 20,position-showcase-d 20,position-showcase-e 20]
+    - [position-showcase-a 33.3,position-showcase-b 33.3,position-showcase-c 33.3]
 
 feature:
   container:
-    - [position-feature-a 20,position-feature-b 20,position-feature-c 20,position-feature-d 20,position-feature-e 20]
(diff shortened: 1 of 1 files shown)


JoomFX:
@Omegatcu well done mate :)

we should also add the `offcanvas` at the bottom with the "Mobile Menu" particle so it can be there by default.

I tried the following codes but I cannot figure out what is the one for the "Mobile Menu" particle:

```
offcanvas:
    - mobile-menu
```

and

```
offcanvas:
    - mobile_menu
```

Omegatcu:
I don't think you can overrwrite the offcanvas position at this time.

JoomFX:
In `_joomla_-_beez3.yaml` and `_joomla_-_protostar.yaml` there is no `offcanvas` added with the "Mobile Menu" particle in it, but somehow this particle gets assigned automatically when you select those Presets. And this is not the case with the G4 Preset you created or with any of the Presets I created :)

It will be best if you guys can shed some light on this @mahagr @w00fz @kevlardev

JoomFX:
@adi8i no Adi, this doesn't work.
with your code (or my code above) you got some strange thing assigned to the Offcanvas Section but it is not the "Mobile Menu" particle, it is some invalid object :)

Omegatcu:
@adi8i your code would create a section with the same name but don't overwrite the section which would be what we need.

Test it yourself - the code would be:
```
offcanavas_section:
  container:
     - [particle-mobile-menu]
```

JoomFX:
This is it @Omegatcu :D

Your code just explains what I was looking for :)

Here is the correct code:

```
offcanvas:
     - [particle-mobile-menu]
```

You do not really need the `container` part here :)

w00fz:
This looks good! I have a couple of things that I would like to be changed before we merge this in.

1. The `drawer` should be the very top section (`drawer` -> `top` -> `header`)
2. The `utility` should go above `feature` and then `breadcrumb` above `utility` (`breadcrumb` -> `utility` -> `feature`)
3. For the Mainbody / Sidebar I would have Mainbody on the left and then just 1 sidebar instead of 2 like you have now

The reason I'm suggesting these changes is because that's more close to how the latest G4 templates have been looking like. You can reference [audacity positions map](http://demo.rockettheme.com/live/joomla/audacity/features/module-positions).

Everything else looks pretty good as is! Good job on that. Please make these changes and I'll merge it in! If you push your changes in your `develop` branch they will show up in here, you don't have to redo the Pull Request.

w00fz:
Thanks you!

mahagr:
In case if someone is wondering what happened, @w00fz reverted the change from wrong branch (release/5.0.0-beta.4) and reapplied it to develop.

PR #131, opened 2015-04-27
Revert "Gantry 4 Preset"

w00fz:
Reverts gantry/gantry5#128


mahagr:
Change was made against release, not develop! 

In case of someone wondered why the change was reverted.

PR #232, opened 2015-05-12
Update spacer.html.twig

WebTread:
This adds the functionality to the spacer.yaml addition we made to add height as an option to the spacer for vertical spacing as well. Good for sidebar spacing or other uses.

### DIFF
--- a/engines/common/nucleus/templates/content/spacer.html.twig
+++ b/engines/common/nucleus/templates/content/spacer.html.twig
@@ -2,5 +2,5 @@
 {% set spacer = gantry.config.getJoined('particles.' ~ segment.type, segment.attributes) %}
 
 {% if enabled and (segment.attributes.enabled is null or segment.attributes.enabled) %}
-    <div class="spacer{{ segment.classes ? ' ' ~ segment.classes|join(' ')|e }}"></div>
+    <div class="spacer{{ segment.classes ? ' ' ~ segment.classes|join(' ')|e }}" style="height:{{ spacer.height }}"></div>
 {% endif %}


w00fz:
Hey @WebTread, I dont think this is necessary. The functionality is already built-in.
If you add a Spacer and you edit its settings, in the Block section you can add the style attribute with the desired height. I just tried it and works fine.

spacer custom style

WebTread:
@w00fz That is great. I think many of us would think Tag is like a Key word like in an article. I had no idea that you could add style in there and then height. Maybe I missed that in documentation. What other Attributes are allowed?

I'm wondering if it would be possible for us to include all the possible Tag Attributes in that tool tip at the very least or a good example to show cause there is nothing there now.

I'm glad this exists. This means my spacer.html.twig should be kicked out as well :(

Now that I got a taste of being a contributor to Gantry5 I will need to find a new way lol

w00fz:
Any attributes `div` allows (that in HTML5 are endless really!) except for `id` and `class` because those have their own fields above. So `style`, `lang`, `style`, `title`, anything really! :)
My favorite one is `data-*` (like `data-my-customattr="test"`) ([see here](http://www.w3.org/TR/html5/dom.html#custom-data-attribute)).

Hah! I'm sorry I didn't merge in your change, it was actually a good thing you brought it up, we should definitely document it but **THANK YOU** for taking the time of putting together a pull request, we do appreciate it very much.

Looking forward for additional contributions!

mahagr:
@FrugalGeek We need to document this one...

PR #233, opened 2015-05-12
Update spacer.yaml

WebTread:
Adds optional Spacer Height field so you can set a Vertical Height as well. Good for Sidebar spacing and other use cases.


w00fz:
See my comment in #232

PR #272, opened 2015-05-15
Gantry 4 Layout with 2 Sidebars

Omegatcu:


w00fz:
I personally think we should just have one generic G4 Layout and with one sidebar only it covers most of the scenarios.

Omegatcu:
as you wish - I need it in a lot of configurations so i thought i share it

PR #284, opened 2015-05-16
Added empty space before "item.class"

JoomFX:
Added an empty space in front of `{{ item.class|e|default('') }}` .
Without this empty space the Custom CSS class gets added at the end of `g-menu-item-link-parent` , for example `g-menu-item-link-parentTEST`


w00fz:
Good catch!

PR #287, opened 2015-05-16
Fullwidth/Boxed Options

JoomFX:
Including Fullwidth and Boxed Layout options in the Admin

PR #288, opened 2015-05-16
Taking the Layout from Admin

JoomFX:
Taking the option the user selects in the Admin, Styles tab, for Fullwidth or Boxed.

### DIFF
--- a/engines/common/nucleus/templates/page.html.twig
+++ b/engines/common/nucleus/templates/page.html.twig
@@ -49,7 +49,7 @@
     {% block page_body -%}
     <body{{ gantry.page.bodyAttributes({'class': [offcanvas_position, 'g-style-' ~ gantry.theme.preset]})|raw }}>
         {{ page_offcanvas|raw }}
-        <div id="g-page-surround">
+        <div id="g-page-surround" class="{{ gantry.config.get('styles.base.layout')|default('fullwidth') }}">
             {% if page_offcanvas|trim %}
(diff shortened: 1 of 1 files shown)


mahagr:
Part of #287

mahagr:
In the future I would rather see a single pull request on simple changes like this. Read Github documentation on how to clone your branch and how to do pull requests from your local branches. It should be easy and the docs in here are the best.

JoomFX:
Thanks @mahagr :)

Just to comment on your suggestions - I had never used Github before the G5 project was released. Since then I'm reading the Github Documentation quite often. You can see my fork of G5 [here](https://github.com/JoomFX/gantry5) . 
I'm now using Sourcetree so I'm doing the Pull Requests at once, for example [this one](https://github.com/gantry/gantry5/pull/294) ;)

mahagr:
@JoomFX We all needed to learn how to work with Github, but I can say you this: once you get it, you will never want to use anything else!

JoomFX:
"*once you get it, you will never want to use anything else!*" - couldn't agree more. I'm already loving it, even though it is still confusing from time to time ;)

mahagr:
The hardest thing in git is to understand that it is a distributed system and not centralized. So you need to be always thinking where you do the change: your local repo / your fork in github / upstream / something else.

w00fz:
Part of #287 and now consolidated in #301

PR #294, opened 2015-05-16
Dropdown Animation

JoomFX:
This is the Dropdown animation feature I posted in Gitter and as per Andy's request I'm opening a Pull Request.

I really wasn't sure how I should apply the changes. I mean, I was wondering if I should delete the `transition` in [_nav.scss](https://github.com/gantry/gantry5/blob/develop/engines/common/nucleus/scss/nucleus/_nav.scss) and also in [nucleus.css](https://github.com/gantry/gantry5/blob/develop/engines/common/nucleus/css-compiled/nucleus.css) instead of overwriting it from my `_dropdownanimations.scss`.


w00fz:
Hey @JoomFX good job on this one, I like it a lot!
I do have a few changes I would like to see before I merge this in, if that is alright with you:

1. We do not use camelCase syntax for class names, `fadeInUp` should be `fade-in-up` for consistency
2. To avoid possible conflicts with third party extensions, we tend to prefix all our classnames. I think it would be safer to prefix the classes for the animations: `g-fade`,  `g-zoom`, `g-fade-in-up`
3. You should add `default: fade` in the YAML, so we fall back to the same behavior as it is now
4. I think the animation should be limited to the first level dropdown if it's of `Extended` type (ex `Fullwidth`)s, the animation for the slides in `Extended` we want to maintain as a sliding in/out effects. Especially with the fade-in-up it doesn't really look good.

JoomFX:
Thanks @w00fz , I'm glad you like it :)

Alright, check it now. The classes are renamed as suggested, I added the `default` in the YAML and I re-wrote the SCSS so it has 2 cases - `.g-standard` and `.g-fullwidth >` , not sure if this is the optimal way so feel free to redo the parts you think can/should be improved.

Also, I wasn't sure what exactly I should do with the Extended/Fullwidth menu when "No Animation" is selected. I removed the `transition` only on the first level dropdown, so the deeper levels still slide in/out :)

w00fz:
Sweet! Gave it a quick look and looks pretty good, I'll merge it in.

Thank you for this, really appreciated. :)

JoomFX:
Niiice :)
Thanks @w00fz !

PR #297, opened 2015-05-17
Add default value to totop particle text

adi8i:
Add default value


w00fz:
Hey @adi8i isn't this already handled [here](https://github.com/gantry/gantry5/blob/develop/engines/common/nucleus/particles/totop.yaml#L23)?  It should automatically pull that value from there, is it not happening?

adi8i:
For now, no, as i remember is a bug there, @mahagr know about this.

mahagr:
Default values aren't used (similar issue to what Joomla has) as I'm not reading blueprints to figure out the defaults.

Regarding to pull request, particle.content|raw|default() makes it not to be raw anymore. You need to have raw as the last filter.

w00fz:
Can you change this @adi8i so I can merge?
Instead of `particle.content|raw |default('To Top')` make it `particle.content|default('To Top')|raw`

PR #301, opened 2015-05-18
Fullwidth/Boxed Options

JoomFX:
Added an option so the user can select between Fullwidth or Boxed Layout from the Admin.


gospelpowered:
Hey guys, how does this work when you have a full width layout by default and want it to be boxed? It takes it down to the g-container width? Isn't the container width controlled in Gantry styles and set to 75rem, which may not be enough for the actual site. Also if you do this the actual section containers will have no padding on the content, i.e. the boxed width will equal the width of the content and that won't look good. 

Or is my thinking wrong here? You want to keep the container separate to the width of the site itself. 

I have implemented a solution where the user has ability to control the width of the page surround, which doesn't affect the container, which works well. See http://www.awesomescreenshot.com/image/235988/b68df3b7340fcc6b85e31c2a0d64add1

JoomFX:
Hey @gospelpowered ,

Probably you expect an answer from the RT guys and hopefully they will also share their opinion on the concerns you mentioned above.

Anyway, here's my opinion... I think the `g-container` approach is the correct one. No, not the correct one - the one I like more :) Also, few weeks ago the RT guys said it in the chat - "*Just put everything in a container*". I did all my G4 Boxed templates the same way - I put everything in `rt-container` which is the equivalent of `g-container` in G4.

Yes, the container width is controlled by Gantry in the Styles tab. And this is actually very nice as you can control the width of your Boxed layout (and the Fullwidth as well).
No, the Section container will have padding and margin because the `g-content` that surrounds each and every element comes with pre-defined margin and padding.
You can also add additional padding to each section. Have a look at the video below, it explains everything:

https://youtu.be/zTfJvK2zVOI

Your approach is very good, unless the user sets the width in `px`. But it can be restricted and the allowed values might be `rem`s and `%` . I actually like your solution, just submit it as a PR. I'm all for making G5 better, so if your approch is better, be it :) Or if the RT guys choose mine, then you can simply modify it and still use yours :D

gospelpowered:
Hey @JoomFX thanks for the response. 

I did test your method and yeah it's a valid way to go. As far as what's best i don't know. I personally think controlling the site width using page surround is nice as it allows the user to then play around with the container width using the breakpoints, i.e. keeping these things separate. So both aspects give the user fine tuned control, they can use rem or % for the boxed width and breakpoints for controlling how close the content is to the walls so to speak. They wouldn't need to then touch the g-content css. 

But i'm new to G5 so i could be totally wrong :)

Any thoughts on this @mahagr @w00fz @kevlardev.

JoomFX:
@gospelpowered yeah, I'm not sure which approach a like more now :D 
Yours has some benefits! I like this *controlling how close the content is to the walls* ;)

I'm also new to G5, we are all new to G5 :D 

So yes, let's wait and see what @mahagr @w00fz and @kevlardev will suggest.

gospelpowered:
Hey @JoomFX hehe in hindsight it was a silly thing to say, G5 did come out recently ;)

I've made a quick video which shows what i mean - https://db.tt/8ak5kNPB

JoomFX:
@gospelpowered yep, I have to admit that yours is more flexible and user-friendly :)
If it does not have any downsides (which I cannot think of any right now), I'd like to see yours integrated in G5 ;)

JoomFX:
@gospelpowered can you submit yours as a Pull Request and we'll link the two so the RT guys can compare and let us know what they think ;)

gospelpowered:
Hey @JoomFX sure thing :)

mahagr:
Competing solution with Pull Request #319 

@kevlardev when you have some time, I'd like you to look into both.

mahagr:
Replaced by per section control in #434

PR #309, opened 2015-05-18
Update totop.html.twig

adi8i:


w00fz:
Thanks

w00fz:
Just a side note, you don't have to delete a pull request and recreate it, you can still make changes in the branch that's been created for you and they will get applied to the same PR you have already opened.

PR #312, opened 2015-05-19
Update logo.html.twig

adi8i:
May be useful to have a hard coded class in some particles. Lets say from aesthetic reasons you want to limit the logo image size in navigation section. Of course you can add a class in particle settings, but this way, that class inadvertently can be removed. 

This is template specific, but i think will be just fine to have that class in there. You have the last word :D .


w00fz:
I think this shouldn't be imposed by us but changed if needed like you did. Really depends on the theme and circumstances.

PR #315, opened 2015-05-19
Saving some memory for echo.

dseguy:


nicktacular:
See https://github.com/serbanghita/Mobile-Detect/pull/403

w00fz:
was just about to start doing the same tests, I've never seen comma concatenation before and by the tests @nicktacular reported I see no benefit whatsoever.

Thank you @nicktacular for pointing that out and good luck @dseguy with your comma concatenation crusade! :smile:

mahagr:
I agree with @w00fz, even if it was better, I would still not accept changes like this. If you can make Gantry to load 5% faster (meaning that you could actually measure the difference), then it would be something worth to consider.

PR #319, opened 2015-05-19
Boxed Width Option

gospelpowered:
Allows the user to enable boxed design and specify in rem or % the width of their site. It leaves the breakpoints setting to control the proximity to the boxed design wall. The user can then control both to a width they prefer.

Not 100% on the mediaqueries.

This is an alternative proposal to #301

### DIFF
--- a/themes/hydrogen/common/blueprints/styles/base.yaml
+++ b/themes/hydrogen/common/blueprints/styles/base.yaml
@@ -21,3 +21,9 @@ form:
       type: input.fonts
       label: Heading Font
       default: 'roboto, sans-serif'
+    width:
+      type: input.text
+      label: Boxed Width
+      description: Set width size in rem or % values
+      default: "75rem"
+      pattern: '\d+(\.\d+){0,1}(rem|%)'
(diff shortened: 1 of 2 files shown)


mahagr:
Competing solution with Pull Request #301

@kevlardev when you have some time, I'd like you to look into both.

adi8i:
you have not forgotten anything?

adi8i:
I am sorry, but i don't see how this code will works ?  
First of all `$base-width` is not assigned to any width, and there is `margin: 0 auto` for `#g-page-surround`, that id doesn't have any `margin` , `margins` are added to `g-container` .

Maybe i don't understand `@if` condition in `sass` is more like a question :)

JoomFX:
@adi8i download both files and test it

`$base-width` is defined in [base.yaml](https://github.com/gospelpowered/gantry5-1/commit/5bf86a52892e44a5787a6ff26f5c57a1c8e0026b) and then the SCSS variable is used in [_core.scss](https://github.com/gospelpowered/gantry5-1/commit/48b987ff1cc150f7bdb4e7bbb9f845eeb3fa61aa). I haven't tested it tho but it should work.

As I told @gospelpowered , his solution seems to be more flexible and user-friendly than mine. The only thing I do not like is that there are quite some adjustments for the different resolutions, while in mine there is just a single class added to the `g-page-surround` element.

I really do not know which one is better. Both have pros and cons ;)

adi8i:
oops didn't seen the width added for large display :laughing:

mahagr:
Related to #434

PR #332, opened 2015-05-21
Adrian full/fixed width

adi8i:
I have my own way to full/boxed page


adi8i:
ehe i need to close this one, will not work fine :)

PR #337, opened 2015-05-22
Update _core.scss

adi8i:


mahagr:
@adi8i Can you explain a bit what you're trying to fix here. Forcing !important doesn't sound nice thing to do and shouldn't be used unless there is high demand on doing it.

adi8i:
@mahagr `.btn-primary` doesn't override default joomla colors, need to be forced. Even if there is `background: $accent-color-1` , still joomla default (blue one) will be used.

mahagr:
Thanks for the explanation, now it's just up to @kevlardev to decide if the change is something he wants to get in. :)

kevlardev:
This is actually an issue thats come up do to some improper ordering of the theme's platform specific styling loading before bootstrap. We will fix this in a different way as it applies not only to the btn-primary but also several other elements.

mahagr:
Fixed in #416

PR #355, opened 2015-05-25
Dropdown Animation default value

JoomFX:
Changing the dropdown animation default class to `g-fade` as `fade` is invalid.


w00fz:
Ah yah I missed that one when I merged your PR before. Thanks.

PR #372, opened 2015-05-27
Update mobile-menu.html.twig

adi8i:

Issue #371, opened 2015-05-27
Ability to define mobile menu trigger point.


marktaylor46:
In addition to the ability to set the breakpoint widths it would be great if we had the ability to define the trigger point at which the menu switches from normal to mobile hamburger/sidepanel. 

I would suggest that an additional field be added to the style tab "breakpoints" that says something like:
"mobile menu threshold point (rems)".

In this way you could still have all the switching of the layout widths but in addition (for example) have the mobile menu show at all viewports sizes, some or none.

It's currently a design fashion to have no "normal" menu at all and instead to just have a mobile menu hamburger and sidepanel to save on screen real estate.

DanGthatRokGallery:
+1
Also it would be nice to implement this for Tablet mode as menu line wrap is a difficult issue to deal with.

adi8i:
+1, right now you can do this, need to override mobile-menu particle. For me is ok like it is now too.

adi8i:
@w00fz now you can close this one too. https://github.com/gantry/gantry5/pull/372 :)

PR #373, opened 2015-05-27
Update breakpoints.yaml

adi8i:


adi8i:
Adding mobile menu breakpoint #372

marktaylor46:
Ok I can see my #371 was closed and merged into #372 and #373 - and it's in Hydrogen - but it's not been put into Callisto (which is where I really wanted it) - So I am going to raised a ticket on RT's internal tracker for Callisto. In my opinion this change should be in all Gantry 5 templates.

adi8i:
That have been added in gantry core so all templates will have this by default. Callisto need only an update.

PR #441, opened 2015-06-09
Default dropdown animation value should not be forced

adi8i:
Default dropdown animation value should not be forced, so anyone can set a default value in `yaml` file.

### DIFF
--- a/engines/common/nucleus/particles/menu.html.twig
+++ b/engines/common/nucleus/particles/menu.html.twig
@@ -97,7 +97,7 @@
     {% if hasCustomWidth and not (root.getDropdown() == 'fullwidth' and item.level > 1) %}
         {% set customWidth = ' style="width: ' ~ item.width|raw ~ '" data-g-item-width="' ~ item.width|raw ~ '"' %}
     {% endif %}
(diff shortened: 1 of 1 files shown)


JoomFX:
It is not force Adi, it must have a default value.
This is what @w00fz asked me to do - https://github.com/gantry/gantry5/pull/294

w00fz:
That `default` isn't forcing the value but just a preemptive fallback in case the animation hasn't been saved in admin.
You can still use `default` in your yaml and that's what it will show up as default value in the dropdown, but unless you save it won't be used and thanks to the `default` in the twig, it will fallback to fade.

This is sort of a known issue we have in Gantry, right now when reading a property that hasn't been saved, doesn't fallback into reading its `default` value set in the yaml.
This is why we are appending the `|default('value')` in the twigs.

At some point we should improve this and cleanup the twigs.

adi8i:
Ok, i know about the issue, i thought it was solved :D. I will try to add a default value in config folder, so the box will be checked by default with the value i want, because i need other class than `.g-fade` as default, i will take a look :) .

mahagr:
Yeah, the issue is still there and I've not had time to look into it yet. As the issue can be avoided by writing better code (hardcoded defaults instead of just failing) and as Joomla has exactly the same issue with its configurations, this issue just isn't top in my list right now.

Issue #577, opened 2015-07-15
Custom JS/CSS atom enhancement


alisamii:
Currently, when you add a Custom JS atom, you can specify the file location and that is about it. However, JSs can be either placed in the `<head>` element or in the footer just before the `</body>` tag. There is even syntax within G5 to insert the script file reference in the head or footer. However, there is no interface within the particle to specify the target insertion point.

It would be very useful to be able to click a checkbox (as in the mockup below) that would insert the script in the footer. The default could be in the `<head>` element. *I'm not sure which option would be better as the default, but in this mockup I have selected default for head and if selected, then insertion would occur in the footer.*

![atom](https://cloud.githubusercontent.com/assets/1498512/8695970/fa5583aa-2ae8-11e5-92c4-045c26a90c17.png)

A further enhancement, which I have not mocked up, is to provide the option to inline the script. This would then inject the contents of the javascript file inside a `<script>` tag rather than link to the external script file.

w00fz:
This was also mentioned in #62, not sure why it was closed.

w00fz:
See @adi8i's contribution in #580, would be nice if you could give it a try. I'll review soon and merge it in later.

w00fz:
This is now implemented, I merged #580.

PR #580, opened 2015-07-15
Javascript atom - improvement

adi8i:
#577 improvement

### DIFF
--- a/engines/common/nucleus/particles/assets.html.twig
+++ b/engines/common/nucleus/particles/assets.html.twig
@@ -18,7 +18,6 @@
 {% block javascript %}
     {% for script in particle.javascript %}
     {% if script.position == false %}
-        {% if script.location %}
         {% set attr_extra = '' %}
         {% if script.extra %}
             {% for attributes in script.extra %}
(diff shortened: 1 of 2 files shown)


w00fz:
Nice!

w00fz:
It just needs to be reviewed, thanks @adi8i, good work!

engines/common/nucleus/particles/assets.html.twig
```diff
@@ -25,7 +26,36 @@
                 {% endfor %}
             {% endfor %}
         {% endif %}
+        
+        <script src="{{ url(script.location) }}" type="text/javascript"{{ attr_extra|raw }}></script>
```
mahagr: This line should be in {% else %} of the next if.

engines/common/nucleus/particles/assets.html.twig
```diff
@@ -25,7 +26,36 @@
...
+        {% set attr_extra = '' %}
+        {% if script.extra %}
+            {% for attributes in script.extra %}
+                {% for key, value in attributes %}
+                    {% set attr_extra = attr_extra ~ ' ' ~ key|e ~ '="' ~ value|e('html_attr') ~ '"' %}
+                {% endfor %}
+            {% endfor %}
+        {% endif %}
+        
         <script src="{{ url(script.location) }}" type="text/javascript"{{ attr_extra|raw }}></script>
```
mahagr: This line should be in {% else %} of the next if.

mahagr:
Actually my comments isn't the full story. I'm just trying to tell that if you have just inline js set and no file, it's still going to add tag for the file.

So the question is how this should work?

engines/common/nucleus/particles/assets.html.twig
alisamii: Does the code, as is, just insert a link to an external script file at the end of the body tag? Or does it inline the script?

Is there any benefit to inlining the script over just linking to the script other than reducing http request(s)?

Best regards,

Ali

w00fz:
Thank you @adi8i this was an awesome PR!

PR #611, opened 2015-07-21
Render dropdown option

JoomFX:
Hey guys,

This is a feature that I need for the website I currently build so I decided to contribute it to our beloved G5 :)

The "Render Dropdown" feature is similar to the "Max Level" option in the "Menu Particle". But with the "Render Dropdown" option you can select for a particular Main Menu item if the dropdown menu should be rendered.

I used to hide the dropdowns by adding a custom CSS class and then "display:none", but the "Render Dropdown" option is a much better approach.


mahagr:
Looks good except it took a bit time to understand what ```loaddropdown=load/noload``` means. :) Personally I would prefer having better name and options for the field, like ```disable_dropdown=0/1``` or even extend using the current Dropdown Style option (maybe renaming it in the process):

    - Use Default -
    Simple
    Extended
    Disabled

JoomFX:
Yah, I also didn't like the names but I could think of better ones :D

Anyway, I just did what you suggested - added the option under "Dropdown Styles". It is actually much better and organized like that. Thanks :)

I also added some more `if`s in order to prevent the loading of some classes that are supposed to be loaded only if there is a dropdown menu ;)

mahagr:
Looks much better now, but I'll sleep overnight as I'm too tired to do code review now.

JoomFX:
No problem Matias :)
Have a rest now and let me know if I need to change/improve something tomorrow.

engines/common/nucleus/particles/menu.html.twig
```diff
@@ -52,7 +52,9 @@
                         {{ SELF.displayTitle(item, gantry) }}
                     </span>
                     {% if (item.children) -%}
```
mahagr: Maybe this instead two nested ifs:

    {% if (item.children) and item.dropdown != 'disabled' -%}

mahagr:
The changes looks clean and nice to me. The only thing I saw was minor thing with 2 nested ifs instead of using just single one. I usually prefer changes that change minimum amount of lines. ;)

JoomFX:
Yep, you are right.
Here you go ;)

w00fz:
Thank you @JoomFX this looks awesome!

JoomFX:
Thanks @w00fz :)

mahagr:
Thanks from the nice addition. I know that people have asked something like this and that for example Kunena had to use its own menu to prevent its menu items from showing up.

JoomFX:
No problem guys... after all it is just few `if`s here and there :)

But yes, I do believe that many Joomla users will use this option. It might not make much sense for WordPress, but for Joomla it does, since you assign modules to menu items, your URLs are based on the menu structure, etc ;)

JoomFX:
@w00fz @mahagr guys, we overlooked something :)

With the current implementation the user is not able to disable second level dropdowns (3rd level menu items) because you are hiding the "Dropdown Style" option for non-first-level menu items.

It will be best if we take the option out of "Dropdown Style", here's an example:

alt

Just let me know if the following YAML is ok, I mean if the names are good and I will submit a new PR:

```
        .disable_dropdown:
          type: select.selectize
          label: Disable Dropdown
          description: Disable the next level dropdown menu.
          default: 0
          options:
            0: No
            1: Yes
```

mahagr:
Hmm.. Can you just create a new issue for this? Otherwise it will get forgotten fast.

PR #664, opened 2015-07-28
Update bootstrap-gantry.css

adi8i:
affix `position: fixed` will be ignored, if section have other position defined. I am not sure this is the right fix, but... I've added left, right values because this is how most of the users use it.


mahagr:
@kevlardev Can you check this one out?

mahagr:
@adi8i Was this fixed somewhere?

adi8i:
No, but my pr is not good, i've added left and right value, bootstrap is not using them by default.

PR #696, opened 2015-08-08
To Top particle minor adjustments

adi8i:
This PR add a font awesome icon picker. I know you can add html in text field, but with this pr you can chose fa icons faster, also the particle have 3 fields so it's not overloaded.

### DIFF
--- a/engines/common/nucleus/particles/totop.html.twig
+++ b/engines/common/nucleus/particles/totop.html.twig
@@ -6,7 +6,7 @@
         <a href="#" id="g-totop" rel="nofollow">
             {% if particle.icon %}<i class="{{particle.icon}}"></i>{% endif %}
             {% if particle.content %}{{particle.content}}{% endif %}
-            {% if not particle.icon and not particle.content %}{{particle.content|default('To Top')|raw}}{% endif %}
+            {% if not particle.icon and not particle.content %}To Top{% endif %}
         </a>
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


mahagr:
@kevlardev @w00fz what do you guys think?

w00fz:
I guess it's ok but obviously like for any other Particle, you can kind of mix and match as you please so a lot of this should be kept simple and the enduser/developer should override the particle to accomplish what they desire.

I'm ok with this addition though, if you can make a few changes. Right now you have it set to have either the Icon or the Text, I'd like to have some logic so that you can end up with any combination:

 1. Icon only (Icon is set, Text is left empty)
 2. Text only (Text is set, Icon is left empty)
 3. Both Icon and Text (both Icon and Text are set)

adi8i:
I can't think of a way to do that, because of this `{{particle.content|default('To Top')|raw}}`, if i'll remove default value, by default the particle will show an empty space and that's not fine.

mahagr:
You can keep the default value, but wrap the content into {% if }

mahagr:
Or better, use something like this:

    {% set content = particle.content ?: (particle.icon ? '' : 'To Top') %}

And use the new variable.

Basically what this does is to use particle.content if it is set, otherwise use default text only if icon isn't there.

mahagr:
Changelog entry is missing!

w00fz:
I'm writing it

PR #783, opened 2015-09-02
Patch 2

adi8i:
Minor adjustments to logo particle

### DIFF
--- a/engines/common/nucleus/particles/logo.html.twig
+++ b/engines/common/nucleus/particles/logo.html.twig
@@ -4,9 +4,9 @@
 <a href="{{ url(particle.url)|default(gantry.siteUrl()) }}" title="{{ particle.text }}" rel="home" class="{{ particle.class }}">
     {% set image = url(particle.image) %}
     {% if image %}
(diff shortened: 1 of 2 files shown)


mahagr:
Not sure if this is the best way to do it as the text kind of needs to be there (do we want to allow empty value on it)?

adi8i:
Hm, then i understood this wrong, it picks the default value from yaml file. But that field can be saved empty, i think this is what you mean.

mahagr:
Well, it's not that I was denying the request; I just wanted to know if the new behavior was what we wanted or not. :)

w00fz:
There isn't a reason for the logo particle to be used if you aren't picking any image nor any  text. So I think the current behavior is the right one. In the PR it's cleaner since we can now use blueprints' default.

Also Adrian, your other change to the url() is a good catch, although you have a typo, you need to remove that extra closing parenthesis at the end, after image.

// Djamil Legato // From iPhone //

mahagr:
If we want no logo to show up, I would just return no content to free up the place particle is taking. In this code you could have empty text, no image and it still contains link.

Issue #906, opened 2015-10-15
Modify Copyright particle to allow links


topwebs:
Links are very standard in the copyright line of websites. For example: © 2015 Cable News Network. Turner Broadcasting System, Inc. All Rights Reserved. contains a link.

Change Copyright particle to allow links, presumably by not escaping html (easiest). I don't really see how  adding link fields will work, as people could use links in a multitude of ways in this particle.

JoomFX:
Just use the "Branding" particle instead of "Copyright" and add your HTML there. For example:

```
© 2015 Cable News Network. <a href="http://companywebsite.com">Turner Broadcasting System, Inc</a>. All Rights Reserved.
```

topwebs:
The Copyright particle has some nice coding for the copyright years which isn't in the Branding particle.

JoomFX:
Here's what you need to do:

1. Create a file called `copyright.html.twig` in `root/templates/g5_hydrogen/custom/engine/particles`. Yes, you need to create the `engine` directory and then create `particles` in it.
2. Paste the following code there:

```
{% extends '@nucleus/partials/particle.html.twig' %}

{% set start_date = particle.date.start|trim in ['now', ''] ? 'now'|date('Y') : particle.date.start|e %}
{% set end_date = particle.date.end|trim in ['now', ''] ? 'now'|date('Y') : particle.date.end|e %}

{% block particle %}
Copyright &copy;
{% if (start_date != end_date) %}{{ start_date|e }} - {% endif %}
{{ end_date|e }}
{{ gantry.platform.filter(particle.owner)|html|raw }}
{% endblock %}

```

3. Use as much HTML in the "Copyright owner" field as you want.

I just tested the above and it works just fine.

I suppose the next thing you will say is that you do not want to override the default particle file - my answer is that this file will never change.

JoomFX:
Or if @adi8i wants he can submit a PR for this.
Just replacing:

```
{{ particle.owner|e }}
```

With this:

```
{{ gantry.platform.filter(particle.owner)|html|raw }}
```

So the HTML will be processed in the "Copyright owner" field.

adi8i:
I think copyright particle is fine like it is now, if everyone will come and ask to modify the files for their own needs, will result in a mess.

We add that and tomorrow will come another user that will ask to remove `html` filter, and so on.

You explained in detail how to create the override, it's a 2 mins job.

JoomFX:
I agree with you @adi8i

[...]

PR #907, opened 2015-10-15
Add Copyright Owner Link

adi8i:
#906


topwebs:
This patch really doesn't solve the issue. If you do not want to use escaped html, then you need to add a text field before the link field and a text field after the link field, in order to cover the ways that most sites use the copyright line. 

I just don't see the issue with using non-escaped html in a particle. You have the custom html particle, why not here?

adi8i:
At least you have an example.

mahagr:
Added wontfix tag as this is already been closed.

PR #908, opened 2015-10-15
Enhance Copyright particle to include html

topwebs:
#906

### DIFF
--- a/engines/common/nucleus/particles/copyright.html.twig
+++ b/engines/common/nucleus/particles/copyright.html.twig
@@ -7,5 +7,5 @@
 Copyright &copy;
 {% if (start_date != end_date) %}{{ start_date|e }} - {% endif %}
 {{ end_date|e }}
-{{ particle.owner|e }}
+{{ particle.owner|raw }}
 {% endblock %}


JoomFX:
Either chage:

```
{{ (particle.filter ? gantry.platform.filter(particle.owner) : particle.owner)|html|raw }}
```

To this:

```
{{ gantry.platform.filter(particle.owner)|html|raw }}
```

Or add the filter option to the "Copyright" particle YAML file.

```
    filter:
      type: input.checkbox
      label: Process shortcodes
      description: Enable shortcode processing / filtering in the content.
```

topwebs:
OK, I've added the filter. So, now the Copyright particle is basically the same as the Custom HTML particle.

JoomFX:
Yep, with the nice code for the copyright years that you want :)

mahagr:
I have troubles of including this change as it breaks existing sites on escaping `&` into `&amp;` which is highly likely string used in copyright text, potentially breaking html in thousands of existing sites.

Maybe add option "Enable HTML" with description off "WARNING: If you enable this feature, you need to escape characters like & => &amp;" and disable HTML by default.

w00fz:
I don't really like this change. 
Filter is quite an expensive method to call (especially in Joomla). It does make sense to have it in the Custom HTML particle, where things can go wild, however in the Copyright particle, the field is controlled and all you will ever do is either link the owner or add some inline tags such as `strong` or `em`.

I think it should just be served as `|raw` to allow you to add the link you were asking, or any other HTML for what matters, but the extra options and filtering, that I feel like it shouldn't be part of Gantry, eventually one could override or create their own Copyright particle, if that's the desired behavior.

This PR should be reverting the `yaml` and `twig`. The `twig` then should just change `{{ particle.owner|e }}` to `{{ particle.owner|raw }}`.

topwebs:
OK. Changed pull request per @w00fz to serve `|raw` in Copyright particle.

w00fz:
Thank you @topwebs !

mahagr:
My concern with copyrights like `Me & You` is still there. Without escaping we break all the existing sites using & in the copyright.

PR #1022, opened 2015-11-23
Fix half missing closing icon on alert

810:
The icon is now half hidden on the alert info messagebox

### DIFF
--- a/engines/joomla/nucleus/css-compiled/joomla.css
+++ b/engines/joomla/nucleus/css-compiled/joomla.css
@@ -516,7 +516,7 @@ th .tooltip-inner {
 
 .alert {
   border-radius: 0.1875rem;
-  padding: 0.938rem;
+  padding: 0.938rem 2rem 0.938rem 0.938rem;
   margin-bottom: 1.5rem;
   text-shadow: none;
 }


w00fz:
Can you verify that the problem applies also to non kunena views, ie `com_content`, and that it's not specific to kunena styling?

Thanks

JoomFX:
@w00fz I can confirm that it is not Kunena specific issue.
I just tested it with a default G5/Hydrogen installation and the issue is reproducible.

I remember this issue when I was starting my first G5 template. I just set `position: static;` to the close button (the "X"). By default it is styled by Bootstrap with `position: relative;`

mahagr:
Fix isn't correct, as its in the compiled file, so we need to re-apply it into proper scss file and recompile.

810:
yeah, I couldn't found the scss file 123, so I pulled the compiled version.

w00fz:
I couldn't merge this PR, targeting a compile CSS and I also didn't quite like the approach you took.
Haven't realized @JoomFX commented but that's exactly how I fixed it now, by removing that `position: relative;` altogether.

Cheers.

PR #1047, opened 2015-12-10
Make decimal point sizes flex compatible

adi8i:
Other sizes have `flex` value too, this one doesn't, will not affect all users, you will notice this only in specific cases. I don't think adding `flex` will adversely affect anything.

Grid With Flex
with-flex

Grid Without Flex
without-flex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     2. [](#improved)
         - Selectize is now name-spaced with a `g-` prefix to avoid potential conflicts
         - Layout Manager: Add Row and Section Settings action icons are now always visible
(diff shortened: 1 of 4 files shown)


w00fz:
We actually never render the the class for the equalized blocks. So if you have 3 elements at 33.3, they will not show the `size-33-3`.

However, I understand you are referencing these classes from your twig? In which case I can see the issue.

adi8i:
Yes i have them in twig, but is not a must, i can just use round values.

mahagr:
I'm fine with this change -- it will make the behaviour more consistent outside of our grid system.

w00fz:
Can you pull my request on your patch? Just wanted to get those few things in before merging.

w00fz:
Cheers @adi8i and nice one :+1:

PR #1067, opened 2015-12-15
Function to call 3-party class and methods

Flowman:
Extend the platform to be able to call any 3-party classes and methods.

Example:
{% set kuser = gantry.platform.method('KunenaUserHelper','getMyself') %}
{{ gantry.platform.method('KunenaRoute','_','index.php?option=com_kunena') }}

### DIFF
--- a/src/platforms/joomla/Framework/Platform.php
+++ b/src/platforms/joomla/Framework/Platform.php
@@ -338,6 +338,20 @@ class Platform extends BasePlatform
         return method_exists($method[0], $method[1]) ? call_user_func_array($method, $args) : null;
     }
 
+    public function method()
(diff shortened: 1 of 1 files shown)


mahagr:
You should already be able to do this by:

    {% set kuser = gantry.platform.call(['KunenaUserHelper', 'getMyself']) %}
    {{ gantry.platform.call(['KunenaRoute', '_'],'index.php?option=com_kunena') }}

Flowman:
I totally missed the other platform file where the call existed.

We can skip this pull request then.

mahagr:
No problem. It was a great idea that I already implemented a few weeks ago. :)

PR #1070, opened 2015-12-17
Small fix. Flex should be `0 value`.

adi8i:


mahagr:
Shrink, not grow. Makes sense.

adi8i:
Sorry for insist on an issue that's not important.  Looks like the code was not compiled.

mahagr:
@w00fz Can you recompile nucleus CSS? I have some issues with the generated CSS.

newkind:
Recompiled ;)

PR #1109, opened 2016-01-12
Mobile menu issues

JoomFX:
Couple of issue with the Mobile Menu when rendered on Desktop.

This video shows the two issues:

https://youtu.be/13UA_W8-CeI


w00fz:
Hey, thanks @JoomFX! I had a quick review and your fixes look good to me. I'd like to have another look in the morning though. 

I am a bit confused on when/how/why this did break since it's been working fine so far and hasn't been really changed in a while.

Also as a side note, the Issue 2 you actually introduce when you fix Issue 1, in fact in the first part of the video, when you show Issue 1, you can see it closing just fine. zIndex does it tho!

JoomFX:
@w00fz Hey Djamil,

No problem at all ;)

In my opinion this issue has always been there. Of course, this is just a feeling, I haven't tested the earlier versions. I think so, because as you can see it is a strange configuration that you/G5 users might have never tested. 
Everything works perfectly fine when you are on Phone, but if you have an Extended Menu together with a Mobile Menu on Desktop version, the `size-xx` class in the Mobile Menu causes the issue :)

And regarding your side note - nope, I tested it many times for this exact reason... to make sure that my first fix does not cause the second issue :smile: Just set the same configuration as in my video, do not apply my fixes and you will be able to reproduce both issues :wink:

PR #1123, opened 2016-01-27
Add flex 100% for size-100

adi8i:
I use `.size-100` in a particle, to render a block full width, it shouldn't be flex too ?  On mobile it goes 100% because are selected all elements that contain class `size`.


mahagr:
I tend to agree that it should.

w00fz:
I think there was a reason for not being flex, but since I can't remember it nor I can see anything wrong with, we can just merge. 😄
Thanks!

PR #1131, opened 2016-02-02
New selectize field that list all pages / posts...

adi8i:
Example:

wp:
  type: wordpress.posts 
  label: Pages
  post_type: page

### DIFF
--- a/platforms/wordpress/gantry5/admin/templates/forms/fields/wordpress/posts.html.twig
+++ b/platforms/wordpress/gantry5/admin/templates/forms/fields/wordpress/posts.html.twig
@@ -2,12 +2,16 @@
 
 {% block global_attributes %}
     {% set post_type = field.post_type|default('page') %}
+    {% set post_status = field.post_status|default('publish') %}
+    {% set post_per_page = field.post_per_page|default('-1') %}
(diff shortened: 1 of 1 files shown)


w00fz:
Nice!

w00fz:
Maybe those other settings can be dynamic as well?

```
'post_status':'publish',
'nopaging':'true',
'orderby':'title',
'order':'asc',
```

adi8i:
Ok, sure.

adi8i:
I changed `nopaging` with `post_per_page` too, if it's dynamic :)

newkind:
I agree! Nice ;)

w00fz:
Sweet, if you ( @adi8i @mahagr @newkind )  don't have anything else to add to this, we can merge it right away :+1:

mahagr:
I'm fine to merge it.

newkind:
I've merged it - thanks @adi8i !

chaladm:
Noticed this was added to the version. How does an end user use this when building particles?

newkind:
You need to set field type to be `type: wordpress.posts` It should render a dropdown in the admin with all posts inside of it.

PR #1134, opened 2016-02-05
Update _offcanvas.scss

Tyrion2737:
Hello,

I just came across this hover.
Is it needed since for example .g-go-back does not have hover?

Olli


mahagr:
@arifinrt Can you take a look?

Tyrion2737:
Sorry! I just realized that .g-menu-parent-indicator gets background, border and hover only if the item links to a page, too to indicate that there are more items 🙈

PR #1162, opened 2016-02-24
Fix for missing <body>

JoomFX:
Fix for missing HTML tags in the new, JS tooltips (description).

### DIFF
--- a/engines/common/nucleus/blueprints/page/body.yaml
+++ b/engines/common/nucleus/blueprints/page/body.yaml
@@ -42,9 +42,9 @@ form:
     body_top:
       type: textarea.textarea
       label: After <body>
-      description: Anything in this field will be appended right after <body>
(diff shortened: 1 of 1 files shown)


newkind:
Maybe we should try with &#60; or &lt;	
 for < and &#62; or &gt; for > ?

JoomFX:
I already tried that... I tried to escape the characters but it still doesn't work.
`&lt;` and `&gt;` are still rendered into `<` and `>` which creates a `<body>` tag and tried to load it.

w00fz:
Does `<code><body></code>` work?

JoomFX:
@w00fz nope, tried that as well :wink:

w00fz:
Ok sounds good to me then! :dancer:

JoomFX:
thanks :wink:

w00fz:
Thank you!

Issue #1161, opened 2016-02-24
Adding stuff after <body> and before </body> from an Atom


JoomFX:
Hey guys,

I'm looking for exactly the same options as the ones under "Page Settings" - `After <body>` and `Before </body>` but for Atoms. I mean, a syntax that will add an element immediately after `<body>`. Is that possible? If not, can it be added in a future release?

This will be a really nice addition. Here's an example...
I'm currently working on a "Page Preloader" Atom. It needs an HTML element immediately after `<body>` in order to work properly. I can add the HTML element from "Page Settings" but imagine how much easier and user-friendly it will be if I can do everything from the Atom. Then the user will just drag & drop the Atom and everything will be working out-of-the-box :wink:

JoomFX:
I tried adding the HTML element with JS but it doesn't work in this case, I mean the element is added but too late as it takes time for the JS to load. The HTML element should be part of the markup by default and therefore I need to add it somehow from the Atom.

If this is not currently possible in G5, can we just add a new "placeholder" in `page.html.twig` similar to `{{ gantry.config.page.body.body_top|raw }}` where we can inject stuff directly from Atoms? Like we can currently add JS and CSS but for HTML elements as well :wink:

mahagr:
I thought we already had issue on this, but... I guess not or it was closed.

JoomFX:
Hey Matias,

Do you think that this can be implemented for the next release? :wink:

mahagr:
I was thinking it already for the yesterdays release, but didn't have time to do it. We'll see. :)

JoomFX:
Thanks Matias :)

JoomFX:
Matias, since you just said in the Gitter chat, that the next G5 release is soon, it is time for me to bug you about this :smile: 

Any chance that this feature can be ready for the next G5 release? I really need to use it for our April's template. Thanks :wink:

JoomFX:
@mahagr @w00fz guys, I know that you are just about to release, but do you think this enhancement can make it for this release or there will be no time for it?

mahagr:
I don't think its going to make it because of I still need to look into some bugs.

JoomFX:
Ok, thanks Matias.
I'll wait for it in some of the future releases :wink:

mahagr:
New tag can be used from anywhere:

```twig
{% pageblock top %}
    -PAGE TOP-
{% endpageblock %}

{% pageblock bottom %}
    -PAGE BOTTOM-
[...]

PR #1163, opened 2016-02-25
Initial works for #1161

JoomFX:
Preparing `page.html.twig` for https://github.com/gantry/gantry5/issues/1161

What should we replace `{# PLACEHOLDER HERE #}` with?

--------

Sorry guys, I guess this PR is not much of a help :disappointed: 
Feel free to just reject it.
I wanted to implement the feature on my own but I guess I still need to learn a lot.

Basically, we just need to replace `{# PLACEHOLDER HERE #}` with something but I cannot figure it out.

### DIFF
--- a/engines/common/nucleus/templates/page.html.twig
+++ b/engines/common/nucleus/templates/page.html.twig
@@ -42,6 +42,22 @@
     {% endblock %}
 {% endset -%}
 
+{# Setting the "After <body>" and "Before </body>" blocks #}
+
+{%- set page_open_body %}
+    {% block page_open_body %}
+        {{ gantry.config.page.body.body_top|raw }}
+        {# PLACEHOLDER HERE #}
+    {% endblock %}
+{% endset -%}
+
+{%- set page_close_body %}
+    {% block page_close_body %}
+        {{ gantry.config.page.body.body_bottom|raw }}
+        {# PLACEHOLDER HERE #}
+    {% endblock %}
+{% endset -%}
+
 {# Head and footer needs to come last because of any of the above blocks may have CSS or JavaScript in them #}
 
 {%- set page_head %}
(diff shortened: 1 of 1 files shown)


mahagr:
Looks good this far and shouldn't really break anything either.

The issue here is that it's really that simple to implement the feature; you would need to do similar stuff that I did for assets. Basically you need to grab twig output from the template and store that into an array until you grab it from there and insert back into the page.

The reason why I haven't already implemented this is that it takes probably a few hours to implement and test the feature.

JoomFX:
I tested it and it works just fine so far... I mean there's no reason not to work since I'm just re-organizing the code, I'm not implementing anything :smile: 

Yes, I tried to implemented the feature by looking into the `assets` and the way you did it. But as soon as I saw the code in `TokenParserAssets.php` I just gave up... this stuff is not for a simple front-end developer :smile: 

I'll just wait for you to implement it, I guess...

And thanks :wink:

mahagr:
Yeah, it was a pain to implement in the first place... But the new code should be somewhat easier as we already have similar classes and functions implemented.

JoomFX:
Great! I hope it will be easier :wink:

w00fz:
Is this getting followed up in a different way? Does the PR need to stay open still? 
Just wondering as we have a bunch of PR stale for a while and I'd like to close the unneeded ones 😄

JoomFX:
@w00fz Hey Djamil,

No, this PR does not need to stay open. There's nothing valueable in it :smile: 

I'm waiting for Matias to implement this feature when he finds some free time... I guess it will be after the Grav implementation.

I haven't said it recently, so thanks a lot for all your hard work guys, it is really much appreciated ;)

mahagr:
I'd vote to close one or the another, no need to have both pull request and an issue for the same thing. 

Will keep the issue open and close this one.

PR #1168, opened 2016-02-29
Update _flex.scss

adi8i:
Use `>` selector to target `.g-content` only for gantry grid, i may reuse the class inside a particle.
Hmm?


adi8i:
Now i am thinking that this code may cause issues for websites already created.

PR #1182, opened 2016-03-07
Fix for Menu Side Offset mixin

JoomFX:
Guys, I don't know why the following code was there, but it was overriding the first `margin` rule and therefore was making the mixin not to work:

```
			> .g-dropdown {
				margin-#{$direction}: 0;
			}
```

### DIFF
--- a/engines/common/nucleus/scss/nucleus/mixins/_nav.scss
+++ b/engines/common/nucleus/scss/nucleus/mixins/_nav.scss
@@ -56,9 +56,6 @@
 				    z-index: -1;
 				}
 			}
-			> .g-dropdown {
-				margin-#{$direction}: 0;
-			}
 		}
 	}
 }


mahagr:
Could you please point out where the first rule is to save us some review time? :)

JoomFX:
Yep :)
The first rule is on line 45:

```
margin-#{$direction}: $offset;
```

mahagr:
@w00fz Do you know why the second rule was added?

w00fz:
Yes, that is meant to be 0 as it refers to the direct dropdown child of a standard menu (`> .g-dropdown`). The offset is only ever applied to the children. This is by design since most of the time you only want to offset the flyouts and keep the first level aligned, the offset behavior is almost never desired to be the same between the two.

JoomFX:
Thanks for the info Djamil.
But I'm quite confused - I need to add a side offset to the first dropdown as well. I can do it without the mixin, but I thought that the two mixins should work the same way... you can add a "Y" offset to the first dropdown, but you cannot add a "X" offset to it... it is just strange and not consistent (in my opinion) :wink:

w00fz:
It's about alignment. If you could add X and y the same way on both, nothing would line up. Not that this might not be what someone would want to do, it's just not considered default behavior by us.

JoomFX:
After re-thinking it I agree that this should be the default behavior.
Sorry for bothering :wink:

w00fz:
So @JoomFX I was thinking about this. What if we add it as a 3rd option for the mixin. By default enabled so it's backward compatible and still uses our intended behavior. However if you decide to use the mixin you can still tweak it to skip that first child margin altogether or tweak it.

```sass
@mixin dropdown-offset-x($offset: 5px, $direction: left, $first_level: true) { // <-- note the 3rd argument
	.g-main-nav {
		.g-standard {
			.g-dropdown {
				margin-#{$direction}: $offset;
				&:after {
					content: "";
				    position: absolute;
				    display: block;
				    top: 0;
				    left: -$offset;
				    right: -$offset;
				    bottom: 0;
				    border-left: $offset solid transparent;
				    border-right: $offset solid transparent;
				    z-index: -1;
				}
			}

            // By default this is enabled
            // You can then pass in `null` or even a different margin if you want, like `5px`
            @if ($first_level == true or type-of($first_level) == number) {
                > .g-dropdown {
                    margin-#{$direction}: if(type-of($first_level) == bool, 0, $first_level);
                }
            }
		}
	}
}
```

This would allow you to now use the mixin by passing as 3rd argument either `null` / `false` or `true` / `numeric value (ie, pixels)`

Examples:
```sass
// default behavior, first-child margin is 0
@include dropdown-offset-x(5px, left); 
@include dropdown-offset-x(5px, left, true);
@include dropdown-offset-x(5px, left, 0px);

// first-child rule for the margin will not render (like your PR)
@include dropdown-offset-x(5px, left, false); 
@include dropdown-offset-x(5px, left, null);

// first-child rule is applied and with a custom margin just for it

// default behavior, first-child margin is 0
@include dropdown-offset-x(5px, left, 10px);
```

What do you think? Also how are you using this exactly by the way?

JoomFX:
Wow, that's really nice and smart Djamil!
Thanks for that :smile: 

I'm currently doing this (the first dropdown x offset) by just adding the margin manually. So it is really not a big deal, but I do believe that this is a very handy addition :)

I'm using this for a design where you have a fixed sidebar/header where the dropdown menu opens to the left/right, not below the menu as it is by default.

So, I'm all in for adding this :wink:

w00fz:
If you'd like to test this (as I haven't) and you see it works both for default behavior and for your use case, let me know and I'd be gladly implementing it.

mahagr:
I reopened pull request -- not because of we are going to merge it, but not to forget the ongoing discussion.

JoomFX:
Guys, I just tested the new mixin and it works perfectly fine!

@w00fz There is just a small typo in one of your examples, you have forgotten to put the quotes on `true`:

```
@include dropdown-offset-x(5px, left, true);
```

Should be:

```
@include dropdown-offset-x(5px, left, 'true');
```

I tested all options (`'true'`, `1` and `10px`) and everything works.
Thanks guys ;)

w00fz:
Sweet, I add it.

Thanks for testing, weird the boolean has to be a string.

w00fz:
This is implemented now.

JoomFX:
Awesome!
Thanks :wink:

PR #1183, opened 2016-03-08
Fix: Menu item that have children is higher

adi8i:
menu
It's not visible with Hydrogen, but you can reproduce it with an empty theme, no style added.

### DIFF
--- a/engines/common/nucleus/scss/nucleus/_nav.scss
+++ b/engines/common/nucleus/scss/nucleus/_nav.scss
@@ -46,6 +46,7 @@
 
             .g-menu-item-container {
                 @include transition(transform .2s ease-out);
+                word-break: break-all;
             }
 
 	    	// Menu Parent Indicator on Main TopLevel


w00fz:
I am not sure about this one. Need @hexplor / @muffindcc / @kat05 to have a look at it and see how this behaves with our templates.

That screenshot you posted, is that an image or an actual menu styled white on white? Also, I take the 2nd element is the only one with children and for that reason it shows taller than the other ones?

adi8i:
Yes, all that boxes are menu items, the second one, is the only one with children. There's only padding, bg white and color white added, anything else comes from gantry.

w00fz:
Thanks :+1: , I've let the other guys know

newkind:
What browser ? Firefox ?

newkind:
Ok, I added a fix for it, but this pull request fix is wrong, so I'm closing it. Let me know @adi8i if it's fine now.

adi8i:
Google Chrome is the browser, line height doesn't solve it.

newkind:
Then I'm unable to replicate it. I tried it with all styling disabled and all is fine.

Firefox regular menu item : https://monosnap.com/file/34g5MmyJzj2BEbzZK1MACkLHSm3H82.png
Firefox parent menu item : https://monosnap.com/file/e7aE5BeEceFcBxWQlB0O4eMJxcs9NG.png

Chrome regular menu item : https://monosnap.com/file/3TW0QcHDhhW63b3oeOEMavXe25p1Jl.png
Chrome parent menu item : https://monosnap.com/file/J8ciAqdhWW0BcAah4bc6Fi6OT8iTeI.png

Heights are the same after my `line-height` fix.

w00fz:
@adi8i can you put a temporary instance up for us to look at?

w00fz:
FYI, this is now properly fixed as you can see in my commit :arrow_up:

w00fz:
FYI, this change had to be reverted because it was causing more problems with the alignments of menu items.

Issue #1213, opened 2016-03-28
Custom section attributes not copied on layout switch


ghost:
If you add custom settings to section( by creating file YOUR_THEME\custom\admin\blueprints\layout\section.yaml), after layout switch and save, values of custom attributes are lost.

mahagr:
There is a reason why section attributes aren't copied over -- some layout presets have their custom attributes for some custom CSS (or looks). By copying attributes, those custom features would stop working.

mmckp:
I don't exactly understand, you say that in layout preset yaml you can define section attributes? I didn't find such possibility in documentation. Maybe code that loads new layout should copy only custom section attributes defined in YOUR_THEME\custom\admin\blueprints\layout\section.yaml ?

mahagr:
Well, layout preset file is identical to any layout file; it can have all the properties + even some more, which cannot be changed from admin.

Its possibility to copy only section attributes when they haven't been set in preset, but that brings another problem as you cannot beforehand know what gets copied and what is already defined in preset. Its more logical either to skip or copy all the attributes...

mmckp:
Hmm I don't know gantry that well, so I can't say if I agree or not with what you are saying, but I have been using Gantry for a while, and such behavior seems like an obvious bug.

mmckp:
Consider scenario, where I add new section "map" in my "Contact" layout. I go to admin panel > Layout, and it displays my old "contact" sections. I want to load my new layout structure, so i click Load>Contact. It reloads my layout, and new "map" section is visible, but all my custom attributes from sections are lost.

mmckp:
Maybe if user selects the same layout from Load dropdown, it should ask whether to reset or copy section attributes value? But it may seem confusing for user, if he doesn't know framework and terminology very well...

mahagr:
I tend to agree, but I also need to keep backwards compatibility with our own themes.

Simplest solution would be to copy attributes only if they had been specified in the current outline. If not, default ones from the new preset/outline would be used. That would allow presets to have their own attributes while still using the ones which user specified.

The issue that comes into my mind is: what if you circle through all the presets before deciding which one to use..?

w00fz:
I believe none of this will matter anymore once 5.3 is released. With the Inheritance in place you can now decide to inherit only the attributes of the section from another outline which means Loading won't override them anymore.

mahagr:
[...]

PR #1214, opened 2016-03-28
Copy section attributes on layout switch

mmckp:
Possible solution to #1213


mahagr:
This change as it is would break some existing behaviour. Let's discuss about the alternative solutions in #1213.

w00fz:
Closing this PR since we just can't merge it as is, discussion can be continued in the issue mentioned above.

PR #1253, opened 2016-04-25
Add emoji font-family

810:


w00fz:
I really think this should be something the theme developer implements rather than being core of Hydrogen. However if it doesn't cause issues I guess we can get this in.

I have questions though:
1. Can you explain why you put the list of emoji fonts after `sans-serif !default` ?
2. I'm really not too familiar with emoji implementation, how do they work? Do they interfere with the stack of fonts before them (roboto, helvetica, etc..).
3. How do they even get picked up since font-family stack works as in first-found-first-use. So if you have roboto installed (which is local so never fails), how can it get to the emoji?
4. Did you test this? Can you post some screenshots/live site of it?
5. Isn't a unicode-range definition required for this to function?

Really missing a lot of explanation here, a description would have been nice :)

Thanks.

810:
1. its needed to just load the font family
2. Its a own font family, that loaded behind the default font family
3. it has loaded beside, there is no conflict with default font.
4. go to: https://newf1.nl/blog
5.  yes you need utf8mb4 database collation.

ps see: http://getemoji.com/

w00fz:
Ok that looks good. So you can mix and match regular text with emoji by just loading the font itself?

Can you please move them before `!default;` I think that's better for SASS

```sass
$base-body-font:				"roboto", "Helvetica", "Tahoma", "Geneva", "Arial", sans-serif, "Apple Color Emoji", "Segoe UI", "Segoe UI Emoji", "Segoe UI Symbol" !default;
$base-heading-font:				"roboto", "Helvetica", "Tahoma", "Geneva", "Arial", sans-serif, "Apple Color Emoji", "Segoe UI", "Segoe UI Emoji", "Segoe UI Symbol" !default;
```

810:
@w00fz I have updated the pr

w00fz:
Thanks! Merged in. I'll recompile the css and add the entry in the Changelog

PR #1261, opened 2016-04-28
Add ability to choose between intro/fulltext image

nikola3244:
Tested and working properly

### DIFF
--- a/engines/joomla/nucleus/particles/contentarray.html.twig
+++ b/engines/joomla/nucleus/particles/contentarray.html.twig
@@ -42,12 +42,16 @@
                     <div class="g-block">
                         <div class="g-content">
                             <div class="g-array-item">
-                                {% if display.image.enabled and article.images.image_intro %}
-                                    <div class="g-array-item-image">
-                                        <a href="{{ article.route }}">
-                                            <img src="{{ url(article.images.image_intro) }}" {{ article.images.image_intro|imagesize|raw }} />
-                                        </a>
-                                    </div>
+                                {% if display.image.enabled %}
+									<div class="g-array-item-image">
+										<a href="{{ article.route }}">
+											{% if display.image.enabled == 'intro' %} 
+												<img src="{{ url(article.images.image_intro) }}" {{ article.images.image_intro|imagesize|raw }} />
+											{% else %}
+												<img src="{{ url(article.images.image_fulltext) }}" {{ article.images.image_fulltext|imagesize|raw }} />
+											{% endif %}
+										</a>
+									</div>
                                 {% endif %}
 
                                 {% if display.title.enabled %}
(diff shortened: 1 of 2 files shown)


engines/joomla/nucleus/particles/contentarray.html.twig
```diff
@@ -42,12 +42,16 @@
                     <div class="g-block">
                         <div class="g-content">
                             <div class="g-array-item">
-                                {% if display.image.enabled and article.images.image_intro %}
-                                    <div class="g-array-item-image">
-                                        <a href="{{ article.route }}">
-                                            <img src="{{ url(article.images.image_intro) }}" {{ article.images.image_intro|imagesize|raw }} />
-                                        </a>
-                                    </div>
+                                {% if display.image.enabled %}
```
w00fz: You should still have a check in here, based on the selected option (`image_intro`, `image_fulltext`) to ensure that the image is available, before going through and render the link + image

engines/joomla/nucleus/particles/contentarray.html.twig
```diff
@@ -42,12 +42,16 @@
...
+									<div class="g-array-item-image">
+										<a href="{{ article.route }}">
+											{% if display.image.enabled == 'intro' %}
```
w00fz: This should probably have an extra check for backward compatibility:

```
{% if display.image.enabled == 'intro' or display.image.enabled == 'show' %}
```

w00fz:
I like this very much, I just left you a couple comments above. If you can fix those we can merge this in!

Thank you!

engines/joomla/nucleus/particles/contentarray.html.twig
nikola3244: Well, I can set that `display.image.enabled == 'show'` is intro image (revert `intro` to `show`), that way we don't need that check

engines/joomla/nucleus/particles/contentarray.html.twig
nikola3244: Not sure what do you mean with
> to ensure that the image is available
To check that URL isn't blank or something else?

w00fz: Yes, if you see how it was before:

```
{% if display.image.enabled and article.images.image_intro %}
```

That was checking if the display image setting was enabled as well as making sure that the image_intro was set (with ` and article.images.image_intro `). In your change however, this check is gone and we need to have it for both `image_intro` and `image_fulltext`, based on whatever setting has been set.

Does it make sense?

engines/joomla/nucleus/particles/contentarray.html.twig
w00fz: I like the `intro` better. Adding that check is fine, it will be just for backward compatibility, and in a future version we can then just remove the check.

w00fz:
Thank you!

nikola3244:
No problem :smile:

Issue #1260, opened 2016-04-28
"Joomla Article" particle limits category selection to 20 categories only


marktaylor46:
A member reported this in forum but I've also confirmed this. The Category dropdown field only shows a maximum of 20 categories in the dropdown. Many user have much more categories than this so the list should not be limited to only 20 categories.

![2016-04-28_15-29-34](https://cloud.githubusercontent.com/assets/11683218/14889278/1886562e-0d56-11e6-95bb-afa3db72393c.png)

http://www.rockettheme.com/forum/joomla-template-photon/258231-joomla-articles-particle-doesn-t-show-full-list-of-categories

PR #1264, opened 2016-05-01
This is the fix for #1260

nikola3244:


nikola3244:
It seems that I have done something wrong with this pull request. It should only change one file

PR #1265, opened 2016-05-01
Fix for #1260

nikola3244:
This small change will fix the problem with only 20 categories showing in Joomla category picker. (#1260)
This added line will override line 43 from `src/platforms/joomla/Joomla/Object/Finder.php`


mahagr:
Limit is there by purpose and you can set the limit from twig, which I totally forgot to do.

nikola3244:
Thanks for the tip @mahagr
Closing this one because of #1266

PR #1310, opened 2016-05-24
Update button.html.twig

### LINKED ISSUE #1308: Button.html.twig bugs
`gantry5/platforms/common/templates/forms/fields/button/button.html.twig`
Bug 1: line 8
Endif, but no starting if. **Fatal Error**
Maybe there was an intention to put something like this? `{% if field.type is defined %}`

Bug 2: line 13
Unneeded closing curly bracket.
currently: `{% if field.icon is defined %} }`
should be: `{% if field.icon is defined %}`

nikola3244:
Fixes #1308


w00fz:
Thank you!

Issue #1313, opened 2016-05-27
Disabled particles in menu editor still render on front end


adi8i:
![1](https://cloud.githubusercontent.com/assets/9744973/15606046/5758f51e-2412-11e6-954b-635679b8a063.png)
![2](https://cloud.githubusercontent.com/assets/9744973/15606096/a0dab8b2-2412-11e6-9f7b-da43c7c2360c.png)

PR #1314, opened 2016-05-27
Hide disabled particles from menu.

adi8i:
I am sure there's a bettter approach I tried to prevent the `<li>` tag from being rendered with no success. 

This PR add `hidden` class to disabled particles, this way hidden class should be modified to `display: none !important`. Or display none could be added inline.
https://github.com/gantry/gantry5/issues/1313


mahagr:
It would be better just not to render disabled particles.

PR #1315, opened 2016-05-27
Don't render menu item when particle disabled

### LINKED ISSUE #1313: Disabled particles in menu editor still render on front end

nikola3244:
Fixes #1313


mahagr:
Much better. What about modules (and widgets in WP)? Did they have toggle and does it work?

w00fz:
It should be fine as Particles, Modules and Widgets should be all of type `particle`. Still would be nice if this was tested on WP as well.

Nice work Nikola!

adi8i:
Works fine with Widgets on Wp, so it should work with Modules too.

Issue #1271, opened 2016-05-05
Updating  WP plugins with G5 activated causes endless Maintenance mode


newkind:
"Briefly unavailable for scheduled maintenance. Check back in a minute."

We need to locate it and fix it - happens everytime when updating things.

mahagr:
I cannot reproduce the issue with WP 3.5.0, didn't try the latest version, though.

newkind:
Can't reproduce anymore. Closing.

Matt-Yorkley:
I'm getting this as well as a recurring issue. The Wordpress debug bar shows this message when updating a plugin hangs in maintenance mode (only shows up under javascript errors, as the update call is done through ajax):

"Error: Syntax error, unrecognized expression: {"error":{"type":"Whoops\\Exception\\ErrorException","message":"Undefined index: type","file":"\/var\/www\/html\/roots\/wp-content\/plugins\/gantry5\/admin\/init.php","line":119}}"

It looks like the gantry5_upgrader_post_install() function in line 116 of /plugins/gantry5/admin/init.php is getting called when other plugins are updated, but the $options array it takes as an argument is empty in this case.

Matt-Yorkley:
After a bit more digging, the $options array that gets checked in line 199 doesn't contain the ['type'] key at all (when other plugins are being updated), only ['plugin'], which contains a string like: "path/to/plugin-file.php".

A quick check at the top of that function (at line 117) fixes the issue:

if( isset($options['plugin']) && $options['plugin'] != "gantry5/gantry5.php" ) { return; }

newkind:
Reopen as this is still the issue.

nikola3244:
Same issue on dev-20916e3ba, WP 4.5.3

Matt-Yorkley:
Have a look at my description of the issue in commit #1365

mahagr:
Should be fixed now, please test it out.

mahagr:
Looks like the issue only happens when `display_errors = On`

PR #1365, opened 2016-06-16
Update init.php

Matt-Yorkley:
Fix plugin updates issue #1271

### DIFF
--- a/platforms/wordpress/gantry5/admin/init.php
+++ b/platforms/wordpress/gantry5/admin/init.php
@@ -114,6 +114,10 @@ function gantry5_upgrader_source_selection($source, $remote_source, $upgrader, $
 }
 
 function gantry5_upgrader_post_install($success, $options, $result) {
+    if(isset($options['plugin']) && $options['plugin'] != "gantry5/gantry5.php") { 
+        return; 
+    }
+    
     if ($success) {
         $theme = isset($options['gantry5_abort']) && !$options['gantry5_abort'];
         $plugin = $options['type'] === 'plugin' && basename($result['destination']) === 'gantry5';
--- a/platforms/wordpress/gantry5/admin/init.php
+++ b/platforms/wordpress/gantry5/admin/init.php
@@ -114,6 +114,7 @@ function gantry5_upgrader_source_selection($source, $remote_source, $upgrader, $
 }
 
 function gantry5_upgrader_post_install($success, $options, $result) {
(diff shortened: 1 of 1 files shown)


mahagr:
The function is not only used for plugin, but also for gantry themes.

Matt-Yorkley:
That's why it checks if $options['plugin'] is set first, then if it is set and $options['plugin'] is not "gantry5/gantry5.php", it returns...

Matt-Yorkley:
The function in question is hooked into the Wordpress filter "upgrader_post_install", which gets called whenever plugins, themes or languages are updated or installed.

The data that gets passed to that function in the $options array as the second argument varies depending on what's being updated, sometimes $options['type'] is set as "plugin" or "theme", sometimes not, sometimes $option['theme'] or $options['plugin'] is set giving the path to the theme or plugin file. As the function gets called on _every_ plugin or theme update, it should have a check to see if the correct plugin is being updated and return if not.

If a plugin is being upgraded, the $options['plugin'] key will always be set (but not if it's a theme or language pack), and if gantry5 is being updated, the $options['plugin'] key will be "gantry5/gantry5.php".

Hence:
if(isset($options['plugin'] && $options['plugin'] != "gantry5/gantry5.php")) {
return;
}

nikola3244:
@mahagr Are you sure that it is used for theme updates on WP? @w00fz said otherwise.
https://gitter.im/gantry/gantry5?at=57736a658c9263ba30253ff9

w00fz:
I think matias meant it's future proof and currently check also theems even though we actually don't use nor check any themes on wordpress since we cannot put any theme in their repository.

mahagr:
Actually it was a note to myself to fix an issue.

The purpose of the above code is to be able to allow user to easily install CI builds and even get back to previous version of Gantry. But the code doesn't do only that: it also allows theme updates to be manually installed instead of using FTP. This second part will fail after this change.

nikola3244:
Yeah, I understand now what you guys mean.
Btw, I can give you a "broken" backup of my testing environment if it can
help with fixing the issue.

mahagr:
The main issue while fixing issues in WP installer has been that we really cannot test the fixes on installer without making a release first. This issue is kind of easier to test, but again, we cannot really test if we break our updates instead.

Anyway, this will be fixed in few days if not today -- its something we cannot ignore.

nikola3244:
Thank you Matias.
It is an important bug.

PR #1381, opened 2016-06-21
Update BuddyPress Assignments.php

horch004:
This update includes the assignments to BuddyPress Member Types and Groups.


newkind:
Looks good! Merged! :)

Issue #886, opened 2015-10-01
Wordpress - qTranslate-X


dazzyon:
I'm using the qTranslate-X plugin to create a multilingual site with Gantry 5.

Translating the following is without problems: Pages, Posts, Menu's, WooCommerce products.

The only issue i have is translating things inside the Gantry 5 particles. With qTranslate-X it's possible to use the following syntax to specify a text for different languages:

[:en]English Text[:de]Deutsch[:]

Even with the Custom HTML particle (process shortcodes enabled) this syntax is ignored (the full code like the example above is visible).

The strange thing is, when i add a widget particle (custom menu) and i set the title with the language tags, it works!
When i do the same with another widget particle, the language tags don't work (full code visible).

mahagr:
@newkind I think this is duplicate to #887

dazzyon:
@mahagr I don't think it's the same issue. The shortcode also doesn't work for Custom HTML particles, where shortcodes are already supported.

dazzyon:
I guess the problem is that the content of the particles doesn't get passed through the Wordpress translation function "__(...)".

When i add the translate(...) function (part of the Gantry framework) to the "filter" function inside Platform.php (the function that enables shortcodes for custom HTML particles if i understand it correctly), the language shortcodes work correctly:

public function filter($text)
    {
        return translate(\do_shortcode($text));
    }

Maybe there is an easy way to make this work for all particles?

bhodeler:
Hi Dazzyon!

I´d like to know if you have got to solve this issue.

I'm doing the translation of a web made with Isotope theme wordpress. I wonder if it is possible to translate the content of the particles with which it works in Gantry 5. My problem is in how to translate texts that I have included in various Gantry particles, because it does not recognize as shortcodes.

Thanks.

dazzyon:
Hi bhodeler,

I don't have a solution for the particles that don't support shortcodes unfortunately.
By editing de Platform.php file as described above it only works for custom HTML particles.

bhodeler:
Ok. Thanks a lot!!!

dazzyon:
@bhodeler 
I created a pull request for the translate function in shortcode enabled particles.

PR #1445, opened 2016-07-13
Update Platform.php

dazzyon:
Passes shortcode enabled particles through the translate function.

See issue #886

### DIFF
--- a/src/platforms/wordpress/Framework/Platform.php
+++ b/src/platforms/wordpress/Framework/Platform.php
@@ -135,7 +135,7 @@ class Platform extends BasePlatform
 
     public function filter($text)
     {
-        return \do_shortcode($text);
+        return translate(\do_shortcode($text));
     }
 
     public function finalize()


w00fz:
@newkind @mahagr any reason why this hasn't been merged in yet? I can merge it in, just want to be sure there isn't anything else involved?

Unless there really isn't a reason to get this merged, I don't see why a PR like this should stay open for this long :)

newkind:
Well I have doubts about it and I want to investigate it a little bit more. I don't think that we should use this translate function as it's an WP internal one and direct calling it should be avoided. I want to find better and more reliable solution firstly and use this if I won't find anything else.

w00fz:
Roger that.

w00fz:
Any progress on this one or should we close the PR?

mahagr:
To be honest, I'm also not liking automatic translation during filtering -- whether to translate a text field or not should be decided by particle. Or just create/use shortcode mechanism to do that.

newkind:
I agree with @mahagr. I think that we should figure out a proper solution for this ie. why it works with widgets and not particles, rather than pass all shortcodes through translate function as this can produce some unexpected results ie. common words like `gallery` could get auto-translated thus making the shortcodes broken.

Let's discuss this in original issue #886 

Closing.

PR #1454, opened 2016-07-15
Fixing minor Modal bug.

coder4life:
Different browsers with padding calculate heights different. Firefox and Chrome have an inconsistency of display here. Chrome respects the 10vh padding-bottom for the g5-dialog box. Firefox calculates heights differently, as such the 10vh is never adding to the bottom of the modal. To force this to render properly, because of how the box-model works with margin properties as being “addition” to heights, we convert the code to margin use. Enforcing this as an “addition” height guarentees the modal will render with the viewport margin (in additon to the modal height based on children).

### DIFF
--- a/platforms/common/scss/admin/_modal.scss
+++ b/platforms/common/scss/admin/_modal.scss
@@ -59,8 +59,7 @@ html.g5-dialog-open {
     }
 
     .g5-dialog.g5-dialog-theme-default {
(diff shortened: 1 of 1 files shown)


mahagr:
Needs to be compiled and checked against Joomla and WP.

w00fz:
Thanks

PR #1503, opened 2016-08-13
Create wpposts.yaml

dudewithamood:
YAML for a Wordpress posts particle - selected posts (rather than from a category)

### DIFF
--- a/engines/wordpress/nucleus/particles/wpposts.yaml
+++ b/engines/wordpress/nucleus/particles/wpposts.yaml
@@ -1,0 +1,171 @@
+name: Post(s)
+description: Display selected WordPress posts and pages
+type: particle
+icon: fa-wordpress
+
+form:
+  fields:
+    enabled:
+      type: input.checkbox
+      label: Enabled
+      description: Globally enable Post(s) particles.
+      default: true
+
+    _tabs:
+      type: container.tabs
+      fields:
+        _tab_posts:
+          label: Posts
+          fields:
+            post.filter.posts:
+              type: wordpress.posts
+              label: Posts
+              description: Select the posts to display.
+              overridable: false
+              post_type: any
+
+            post.display.image.enabled:
+              type: input.checkbox
+              label: Show image
+              description: Select if the featured image of the post should be shown.
+              default: true
+
+            post.display.title.enabled:
+              type: input.checkbox
+              label: Show title
+              description: Select if the post title should be shown.
+              default: true
+
+            post.display.title.limit:
+              type: input.text
+              label: Title Limit
+              description: Enter the maximum number of characters the post title should be limited to.
+              pattern: '\d+(\.\d+){0,1}'
+
+            post.display.text.type:
+              type: select.select
+              label: Content Type
+              description: Select if and how the article text should be shown.
+              default: content
+              options:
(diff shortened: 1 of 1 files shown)


newkind:
FIrst of all thanks for the commit. Second thing is that I don't think we should have a separate particle for this - if we would want to integrate it, it should be part of the existing particle. This also means that we would have the same functionallity not only for WordPress but also for Joomla as they need to be as similar as possible to each other.

dudewithamood:
The existing particle being contentarray? Or have I missed something? I understood contentarray was also Wordpress specific but agree that ideally one particle should work across all platforms. When I looked into this I found two field templates under plugins/gantry5/admin/templates/forms/fields/wordpress - I haven't seen anything that does the equivalent for Joomla (or Grav).

Perhaps as an alternative there's a GitHub project set up for user contributions that sit separate from Gantry? At least that way they can be shared and maintained.

Should I close this one? This is my first contribution to a project (other than logging bugs and requesting features) so I'm new to it and not sure exactly how it works.

dudewithamood:
I see there's a contentarray for Joomla also. Hadn't realised as I only ever look at the Wordpress code.

Still not sure what you mean by existing particle... happy to try and combine. Since creating this it's my new favorite particle - I'd been missing it and now use it extensively.

newkind:
Yah, Joomla also has one ;)

I mean that having two separate particles - one for posts/articles from Categories and another one for selected posts/articles is not a good idea. Both functionalities should be combined into already existing particle.

Also there's another issue with big sites - how would you select a page or post if site has ie. 10 000 posts ? People tend to have such massive amount of content and having a huge selectbox with 10 000 items is clearly not a way to go ;)

Don't get me wrong, I think it's an awesome functionality and tons of people will find it useful, we just need to think this through and choose a proper aproach to avoid any possible issues.

Thanks for you hard work! :)

dudewithamood:
I started trying to do it as one but wasn't sure how to cleanly handle where fields were only relevant to category. If there was a way to select between post or category, and then selectively reveal fields?

I agree on the huge select box. I quickly took it into a site where I have a lot of content and the one think I thought was that if the select was to only work based on typing using AJAX (rather than scrolling) then the list wouldn't need to load more than a few that are visible.

w00fz:
What's the decision here? Does the PR need some work still or is it something that needs to be discussed in an issue and we can close this PR?

dudewithamood:
A few considerations:

1. I see the benefit of this being part of contentarray (rather than a separate particle) as there is a large amount of duplication - should I submit a feature request for this detailing considerations such as the large select field?

2. The more I'm advancing my own particles the more I see benefit in fields (and tabs) only showing dependent on selections in other fields. For example, I now have an entire tab for featured image with various settings that control how the image is displayed... obviously if the featured image is not being displayed all these fields are not needed. If nothing currently existing for that I can submit a separate feature request.

3. For those such as myself that are writing particles that would not necessarily form part of Gantry, what is the best way to share these in GitHub so others can use and contribute to them?

Otherwise I'm happy for this to be closed - my own version has already mutated into something far more advanced.

newkind:
I'm closing this PR as this needs some more work. Let's move this discussion into #1541

PR #1523, opened 2016-08-25
Fix toggle.toggle

nikola3244:
As I [wrote on Gitter](https://gitter.im/gantry/gantry5?at=57bee8dcf066bd731b494c4f), toggle wasn't working with this YAML:
```
toggle_element:
  type: toggle.toggle
  label: Test toggle
```
Result is: Just label "Test toggle". If we inspect HTML, output is:
```
<div class="switch-toggle switch switch-0">
        </div>
```

By looking at previous code, 0 from `switch-0` means that there aren't any items in `field-options`. So, I tried this:

```
toggle_element:
  type: toggle.toggle
  label: Test toggle
  options:
    key: test
    name: test
```
Which worked! But the output was broken:
<kbd>
screenshot_2016-08-25_14-56-55
</kbd>

So, I went on fixing the code and you can see the result. A normal working toggle:
<kbd>
screenshot_2016-08-25_14-35-48
</kbd>

This YAML now works perfectly with it.
```
toggle_element:
  type: toggle.toggle
  label: Test toggle
```

As you can see, I removed the foreach loop, but if it's important someone can bring it back.
Output value has been checked, it returns true and false when supposed to.

Something that I forgot to do is `default: true|false` and maybe even `required: true|false`.

### DIFF
--- a/platforms/common/templates/forms/fields/toggle/toggle.html.twig
+++ b/platforms/common/templates/forms/fields/toggle/toggle.html.twig
@@ -1,23 +1,8 @@
 {% extends 'forms/' ~ layout|default('field') ~ '.html.twig' %}
 
-{% set value = (value is same as(false) ? 0 : value) %}
-
 {% block input %}
-    <div class="switch-toggle switch switch-{{ field.options|length }}">
-    {% for key, text in field.options %}
-        {% set id = "toggle_" ~ name ~ key %}
-        <input type="radio"
-               value="{{ key }}"
-               id="{{ id }}"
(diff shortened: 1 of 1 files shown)


mahagr:
Cannot you just use `enable.enable` for this?

```
toggle_element:
  type: enable.enable
  label: Test toggle
```

Toggle was meant for multi-toggle with text in the options. I agree that it may be broken, but its for different use (Yes | No | Use Default) than enable field.

nikola3244:
How did I miss that? :smile: 

Nevermind then, I'll close this one.

PR #1545, opened 2016-09-02
Update posts.html.twig

dudewithamood:
Changed so that only actual posts can be selected.

### DIFF
--- a/platforms/wordpress/gantry5/admin/templates/forms/fields/wordpress/posts.html.twig
+++ b/platforms/wordpress/gantry5/admin/templates/forms/fields/wordpress/posts.html.twig
@@ -16,11 +16,15 @@
     {% set posts = gantry.platform.call('Timber::get_posts', array) %}
     {% set Options = field.selectize.Options %}
     {% set options = [] %}
+    {% if posts %}
         {% for id,post in posts %}
             {% set id = post.id %}
             {% set options = options|merge([{value: id, text: post.post_title}])  %}
         {% endfor %}
 
         {% set field = field|merge(field.selectize.Options|default({}))|merge({selectize: { Options: options }}) %}
+    {% endif %}
+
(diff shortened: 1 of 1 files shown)


platforms/wordpress/gantry5/admin/templates/forms/fields/wordpress/posts.html.twig
```diff
@@ -16,11 +16,15 @@
...
+    {% if posts %}
         {% for id,post in posts %}
             {% set id = post.id %}
             {% set options = options|merge([{value: id, text: post.post_title}])  %}
         {% endfor %}
 
         {% set field = field|merge(field.selectize.Options|default({}))|merge({selectize: { Options: options }}) %}
+    {% endif %}
+
+    data-selectize="{{ (field.selectize is defined ? field.selectize|json_encode()|e('html_attr') : '') }}"
```
w00fz: I don't think this is needed? `posts` already extends the `selectize` field which adds the data-set on its own.

adi8i: Without data-set you can type in anything, in this case the input should allow you to input only available posts. Same as wordpress.categories.

w00fz: Then you probably just need to change line 25 and force `create` to false?

```
{% set field = field|merge(field.selectize.Options|default({}))|merge({selectize: { Options: options, create: false }}) %}
```

adi8i: Ok, you are the master here.

w00fz: That works?

adi8i: Yup

PR #1546, opened 2016-09-02
Extend the field to allow selecting custom taxonomies

adi8i:


dudewithamood:
Okay, I see you're working on this... I have a slightly different take on it...
```
    {% set array = {
        'public': true,
        '_builtin': false,
    } %}
    {% set taxonomy = gantry.platform.call('get_taxonomies') %}
    {% set orderby = field.orderby|default('name') %}
    {% set order = field.order|default('asc') %}
    {% set hide_empty = field.hide_empty|default(false) %}
    {% set number = field.number|default(0) %}
    {% set hierarchical = field.hierarchical|default(true) %}
    {% set array = {
        'taxonomy': taxonomy,
        'orderby': orderby,
        'order': order,
        'hide_empty': hide_empty,
        'number': number,
        'hierarchical': hierarchical,
    } %}
    {% set categories = gantry.platform.getCategories(array) %}
```
Feel thought that taxonomy should exclude nav_menu, link_category and post_format

adi8i:
Sorry, I don't fully understand what you mean. It's a `category` field, here i extend it to allow custom taxonomies (category is tax), first of all i had in mind woocommerce product categories, wordpress have lot of available options, that doesn't mean we have to add them all. 

It have to be backward compatible. More than this it's an admin field, most of the options added here are not reflected on front end (correct me if I am wrong). For example order field, you tell the field to show you the categories ordered ASC or DESC, it doesn't order anything on front end, so i don't see how useful most of the options will be.

dudewithamood:
Yes, a category field, exactly... the main difference is that there's no need to specify which taxonomy as **all** taxonomies are included... perhaps an improvement would be a combination such that where field.taxonomy == 'all' it uses get_taxonomies to fetch all the taxonomies rather than being limited to the one specified.

Would it be possible for field.taxonomy to be an array so in the YAML config for the field there could be one taxonomy, more than one, or 'all'.

And yes, it's an admin field and has no bearing on what's displayed in the front end. It sole purpose is to select one or more categories from which posts will be fetched.

If you like I can PM you a link to a site that has an example.

dudewithamood:
How's this? It's working for me with the exception that I'd like to remove nav_menu, link_category and post_format from the taxonomy array but don't know how to do that.

```
    {% if field.taxonomy == 'all' %}
      {% set array = {
          'public': true,
      } %}
      {% set taxonomy = gantry.platform.call('get_taxonomies', array) %}
    {% else %}
      {% set taxonomy = field.taxonomy|default('category') %}
    {% endif %}
    {% set orderby = field.orderby|default('name') %}
    {% set order = field.order|default('asc') %}
    {% set hide_empty = field.hide_empty|default(false) %}
    {% set hierarchical = field.hierarchical|default(true) %}
    {% set array = {
        'taxonomy': taxonomy,
        'orderby': orderby,
        'order': order,
        'hide_empty': hide_empty,
        'hierarchical': hierarchical,
    } %}
    {% set categories = gantry.platform.call('Timber::get_terms', array) %}
```

w00fz:
I don't mind having these extra options for sorting and such, as long as they are just yaml parameters a developer can specify but don't appear in the actual admin field as well as if they have good defaults.

I can see that being useful in some cases where one might want to display the list ordered by X sorted desc and in another field they might want to show the list order by Y and sorted asc.

So as long as they are developers only options, i think they are fine to have.

dudewithamood:
Yes, just yaml params for devs. The defaults are the standard defaults for those function in WP.

Here's an update to the first part to exclude the built in taxonomies that wouldn't make sense in this context:
```
    {% if field.taxonomy == 'all' %}
      {% set array = {
          'public': true,
          '_builtin': false,
      } %}
      {% set taxonomy = gantry.platform.call('get_taxonomies', array)|merge(['category']) %}
    {% else %}
      {% set taxonomy = field.taxonomy|default('category') %}
    {% endif %}
```

w00fz:
Can we rename the `array` variable to something more meaningful? ie, `options` / `settings`

w00fz:
I think @adi8i's PR is ok to merge, althoug does it make sense to expand also to @dudewithamood's proposal @newkind @mahagr ?

dudewithamood:
As I've been able to make the change under the custom path I'm not that attached - the change does what I need it to and I can carry on with what I'm doing.

Having said that, two points:

1. It is backward compatible and will not affect any existing implementation of that field.
2. In contentarray where it's used this change makes that particle incredibly versatile since most things in WP are posts and quite often (actually every site I'm working on) I have use of it without needing the custom fields a third party or custom post type may have i.e. just one or a combo of title, image, content and url. Some examples where this one particle meets all needs (with that change implemented):
  - A list of latest products from a product category in WooCommerce
  - A list of guest house rooms where a room is a custom post type

Where I do need to use custom fields I'll write a specific particle or use the Custom HTML particle - that's for 20% of the requirement... 80% can be met with contentarray.

I feel this can be of great benefit to template developers and help to bridge the gap between less technical users and developers.

newkind:
@adi8i Can you add changelog entry to your PR ?

I don't mind extra options to be there, however I feel that if they are there - they should be documented so devs would be aware of them.

I think that we shouldn't list all taxonomies items at once as let's remember that this selectize field doesn't display the existing taxonomies list but taxonomy **items** list, so we would end up with tons of items that would be duplicating themselves without possibility to check which one belongs to categories, tags or any other.

Opinions ? If you guys agree and @adi8i will introduce changes I mentioned above, I'm fine with this PR.

adi8i:
I added it. Thanks @newkind you made a better point, which I fully agree, in more complex projects you'll probably handle this requests by twig rather than yaml fields.

PR #1567, opened 2016-09-08
Fixed xml errors outside parseAssetsFunc triggering dealXmlError

Chrissi2812:
if an DOM-Error occured before **parseAssetsFunc** is called the **dealXmlError** shows a error with wrong information.


mahagr:
Thanks for the fix!

Issue #1541, opened 2016-08-30
Content Array particle should allow to select single posts/articles


newkind:
This issue is about PR #1503 

This needs some more discussion and simultanious support for all platforms.

dudewithamood:
An option: a select that works with typing using AJAX (rather than scrolling) then the list wouldn't need to load more than a few that are visible e.g. once three letters are typed an AJAX search is performed that returns 5 results... the same way that 'Search for plugins' is now working in the latest version of Wordpress.

JoomFX:
My suggestion is to make the picker (the YAML field) as a button that triggers a modal window. Actually, it should be an input field so the user can type in (and remove) the article IDs manually, but next to the input field there should be a button that will trigger a modal window which will show all the articles with pagination and filter options.

This modal windows will actually be the one from the platform. I don't really know how it is in WP and Grav so I will explain how I imagine the `joomla.articles` YAML field type.
Once you click on the button that I described above, the Joomla's modal window with the articles opens. Basically, the idea is absolutely the same as in creating a "Single Article" menu item or as you already did it in RokSprocket. [This video](https://youtu.be/-EUIqsmWZWE) explains what I mean :wink: 

If possible, the `joomla.articles` YAML field type should fallback to an `input.text` field in older G5 versions.

adi8i:
In wordpress an workaround is to input post ID, not too user friendly anyway.

JoomFX:
@adi8i @newkind Guys, have a look at how I did it in Joomla - https://github.com/gantry/gantry5/pull/1591

It will be great if one of you can add this for WP so we can keep the "Content Array" particle relatively consistent across the different platforms.

Adi, entering the post ID manually should be fine for now... until a proper YAML field for each platform is added.

JoomFX:
Hey guys,

I know that you have other priorities, but do you have any update about this - a "Select Single Article/Post/Page" YAML filed type?
What are your plans about it? Do you still plan to implement it?

AmibeWebsites:
I use this code in a file <my theme>/admin/templates/forms/fields/wordpress/posts.html.twig
```
{% extends "forms/fields/input/selectize.html.twig" %}

{% block global_attributes %}
    {% set post_type = field.post_type|default('page')|split(' ') %}
    {% set post_status = field.post_status|default('publish') %}
    {% set posts_per_page = field.posts_per_page|default('-1') %}
    {% set orderby = field.orderby|default('title') %}
[...]

PR #1591, opened 2016-09-25
Adds the ability to select single articles by entering their IDs.

JoomFX:
It would also be nice to add a `pattern` for this YAML field, but I couldn't figure out the correct RegEx.

But it's not really needed because in the near future this `input.text` field should be replaced by the (not yet available) `joomla.articles` YAML field type (see https://github.com/gantry/gantry5/issues/1541 for more deatils).

### DIFF
--- a/engines/joomla/nucleus/particles/contentarray.html.twig
+++ b/engines/joomla/nucleus/particles/contentarray.html.twig
@@ -22,7 +22,7 @@
 
     {# Content Finder #}
     {% if filter.articles %}
-        {% set article_options = filter.articles ? {id: [filter.articles|split(',')]} : {} %}
+        {% set article_options = filter.articles ? {id: [filter.articles|replace(' ', '')|split(',')]} : {} %}
         {% set article_finder = joomla.finder('content', article_options).published(1).language() %}
     {% else %}
         {% set article_finder = joomla.finder('content').category(categories).published(1).language() %}
--- a/engines/joomla/nucleus/particles/contentarray.html.twig
+++ b/engines/joomla/nucleus/particles/contentarray.html.twig
@@ -21,7 +21,12 @@
     {% set categories = joomla.finder('category', category_options).published(1).language().limit(0).find() %}
 
     {# Content Finder #}
-    {% set article_finder = joomla.finder('content').category(categories).published(1).language() %}
+    {% if filter.articles %}
+        {% set article_options = filter.articles ? {id: [filter.articles|split(',')]} : {} %}
+        {% set article_finder = joomla.finder('content', article_options).published(1).language() %}
+    {% else %}
+        {% set article_finder = joomla.finder('content').category(categories).published(1).language() %}
+    {% endif %}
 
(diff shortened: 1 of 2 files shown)


engines/joomla/nucleus/particles/contentarray.html.twig
```diff
@@ -21,7 +21,12 @@
     {% set categories = joomla.finder('category', category_options).published(1).language().limit(0).find() %}
 
     {# Content Finder #}
-    {% set article_finder = joomla.finder('content').category(categories).published(1).language() %}
+    {% if filter.articles %}
+        {% set article_options = filter.articles ? {id: [filter.articles|split(',')]} : {} %}
```
w00fz: Can you change

```twig
[filter.articles|split(',')]
```
to

```twig
[filter.articles|replace(' ', '')|split(',')]
```

This way is less prone to user mistakes when dealing with spaces and more future proof as now you can use `1,2, 5 , 6, 10` which will always be translated to `1,2,5,6,10`.

JoomFX: Hey Djamil,

Yeah, it's not a problem to apply the change you suggested.
But I already tested it (the current/original version in the PR) with all of the following and it works just fine:

```
1,2,5,6,10
```

```
1, 2,5,6 ,10
```

```
1,2,5,6,10, (notice the comma at the end)
```

```
1 , 2 , 5 , 6 , 10
```

Basically, I'm doing the same thing that Matias did with the "Categories":

```
{% set category_options = filter.categories ? {id: [filter.categories|split(','), 0]} : {} %}
```

So yeah, just let me know if you still want me to apply the change you suggested and I'll be happy to do it :wink:

w00fz: Oh I know it's already being parsed and cleaned up at a lower level, but I think it's best to remove the spaces via twig and going with the assumption that we don't really know if they will or will not get removed to whatever final method we pass the data to.
I believe this is just a better practice: pass the data the way you know it should be handled as. This is what I meant by future proof, in the eventuality that for whatever reason the end point is not stripping spaces, this implementation will most likely fail, however if we had it set to never pass spaces in the first place, we know we will be ok.

And I know I'm being picky here, it's just a minor enhancements that I'd like to see added as I know it doesn't  costs much in terms of performance. 😃

JoomFX: Alright, I see now.
Thanks for the explanation :wink: 

I just committed the changes you requested.

mahagr: In joomla the Finder class is already filtering the input.. So you can pass whatever you want into it and it should remove all bad input.

w00fz: I think they will get implemented in WP and Grav as they gets requested.

Thanks for the change.

w00fz:
We need the CHANGELOG entry.

PR #1608, opened 2016-10-07
Fixed: Articles in nested categories show the whole category tree instead of just the direct category.

JoomFX:
Currently, the "Joomla Articles" particle shows the whole category tree for articles in nested categories:

http://i.imgur.com/KPTicr4.png

I do believe that this is not what most users will expect.

### DIFF
--- a/engines/joomla/nucleus/particles/contentarray.html.twig
+++ b/engines/joomla/nucleus/particles/contentarray.html.twig
@@ -96,17 +96,14 @@
                                         {% if display.category.enabled %}
                                             {% set category_link = display.category.enabled == 'link' %}
                                             <span class="g-array-item-category">
-                                                {% for category in article.categories %}
-                                                    {% if loop.last %}
-                                                        {% if category_link %}
-                                                            <a href="{{ category.route }}">
-                                                                <i class="fa fa-folder-open"></i>{{ category.title }}
-                                                            </a>
-                                                        {% else %}
-                                                            <i class="fa fa-folder-open"></i>{{ category.title }}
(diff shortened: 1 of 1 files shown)


engines/joomla/nucleus/particles/contentarray.html.twig
```diff
@@ -97,12 +97,14 @@
                                             {% set category_link = display.category.enabled == 'link' %}
                                             <span class="g-array-item-category">
                                                 {% for category in article.categories %}
-                                                    {% if category_link %}
-                                                        <a href="{{ category.route }}">
+                                                    {% if loop.last %}
```
w00fz: Why not `{% for category in article.categories|last %}` instead of looping through all and check for `loop.last`?

JoomFX: Hey Djamil,

The first thing I tried was actually `{% for category in article.categories|last %}` but it doesn't work, no category is shown at all.

I'm not sure I understand your second comment. Could you please paste the code as you want it to be?

The only way I could make it work was the one I committed 😉

w00fz: This is what I meant

```twig
{% set cat = article.categories|last %}
```

So no loop is needed.

JoomFX: Thanks 😉 
Have a look now.

w00fz:
Cheers @JoomFX.

/cc @w00fz: Add CHANGELOG

PR #1610, opened 2016-10-09
Make settings link appear under Settings menu (fixes issues with multisite)

dudewithamood:
Adds 'manage-options' permissions check and fixes issue on Multisite installs where settings cannot be accessed due to insufficient permissions

### DIFF
--- a/platforms/wordpress/gantry5/admin/settings.php
+++ b/platforms/wordpress/gantry5/admin/settings.php
@@ -8,23 +8,26 @@ add_filter( 'plugin_action_links', 'gantry5_modify_plugin_action_links', 10, 2 )
 add_filter( 'network_admin_plugin_action_links', 'gantry5_modify_plugin_action_links', 10, 2 );
 
 function gantry5_register_admin_settings() {
-    register_setting( 'gantry5_plugin_options', 'gantry5_plugin' );
+    if ( current_user_can( 'manage_options' ) ){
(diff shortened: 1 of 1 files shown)


mahagr:
The change looks good, its just missing a changelog entry (I can do that in a bit).

newkind:
Is there any reason why you think that we should have this extra menu item under "Settings" ? :) Just want to hear your opinion ;)

dudewithamood:
Hi @newkind and @mahagr - I've been away for two weeks so only seeing this now.

One of the challenges I have with WordPress is its extreme flexibility - not WordPress per se but the lack of standardization building on top of it that can result from this flexibility.

I've personally felt confused about where the Gantry settings are, and that getting to then via the Gantry page was like a back door and I didn't know where the front door was. Generally other plugins I use either have their own settings link as a sub-menu, or there is a sub-menu link under the Settings menu. IMHO this feels more standard.

Also (I believe - could be wrong), having it as a menu item provides some additional control over which roles can access the settings.

mahagr:
No worries, I was on my vacation as well. :)

I'll talk with @newkind about this change; personally I am not against having the menu item there, but I'm not WP expert either.

w00fz:
I'm ok with this change too, can we merge it in please?

faithbible:
I can confirm that this is an issue and that the file changes resolve it.

mahagr:
Sorry, totally forgot to merge this before release!

Issue #1607, opened 2016-10-07
logo/image particle logo.html.twig always adds a link


marktaylor46:
Suggested improvement:

Add an checkbox option to say "no link" in the particle 

If you are using this particle to just display an image then it is often the case that you won't want a link at all. Note that it would not be sufficient to check for a link value as (correctly) the particle default to the "home" link if no value is entered. 

As a "logo" particle the functionality would be fine as it is but the actual description of the particle is logo/image particle.

hacki65:
Good idea 👍

w00fz:
PR welcome for this.

dudewithamood:
@marktaylor46 you can override the particle by editing logo.yaml and logo.html.yaml, and saving them under <your theme>/custom/particles - read http://docs.gantry.org/gantry5/advanced/creating-a-new-particle and http://docs.gantry.org/gantry5/advanced/particle-yaml-field-types, bearing in mind it doesn't need to be a new particle as it will override the particle when you keep the file names the same as the existing one.

This is what I've done to meet my own requirements for this particle.

I'm wondering if it would be useful to have an entirely separate project for user contributed particles and particle changes that aren't necessarily part of Gantry. That way people can pick and choose the ones they want, and contribute changes without it being tied into the Gantry framework as such.

marktaylor46:
@dudewithamood Thank you, yes, I am fully aware how to create overrides for particles (I am MrT moderator on the RT forums) but I believe this change would benefit many people and that's why I raised the issue so that Gantry 5 could be changed. I'll get around to the PR in the next few days.

dudewithamood:
Ah, okay, fair enough - I don't really know who's who...

I have also used this particle for placing images so in that sense, yes I agree.

PR #1611, opened 2016-10-09
Logo Particle: Add option to disable url

adi8i:
If you like the approach :). #1607

### DIFF
--- a/engines/common/nucleus/particles/logo.yaml
+++ b/engines/common/nucleus/particles/logo.yaml
@@ -21,10 +21,10 @@ form:
       label: Image
       description: Select desired logo image.
       
-    url_disable:
+    nolink:
       type: input.checkbox
-      label: Disable Url
-      description: Disable image url.
+      label: No Link
+      description: Renders Logo/Image without any link.
       default: false
 
     svg:
--- a/engines/common/nucleus/particles/logo.yaml
+++ b/engines/common/nucleus/particles/logo.yaml
@@ -20,6 +20,12 @@ form:
       type: input.imagepicker
       label: Image
       description: Select desired logo image.
+      
+    url_disable:
+      type: input.checkbox
+      label: Disable Url
+      description: Disable image url.
+      default: false
 
     svg:
       type: textarea.textarea
--- a/engines/common/nucleus/particles/logo.html.twig
+++ b/engines/common/nucleus/particles/logo.html.twig
@@ -8,9 +8,7 @@
     
     {% if particle.url_disable == false %}
         <a href="{{ url }}" title="{{ particle.text }}" {{ rel|default('')|raw }} {{ class|default('')|raw }}>
-    {% else %}
-        <div {{ class|default('')|raw }}>
-    {% endif %}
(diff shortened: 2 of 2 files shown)


engines/common/nucleus/particles/logo.yaml
```diff
@@ -20,6 +20,12 @@ form:
       type: input.imagepicker
       label: Image
       description: Select desired logo image.
+      
+    url_disable:
+      type: input.checkbox
+      label: Disable Url
+      description: Disable image url.
```
w00fz: Can you change `Url` to `URL`. Or perhaps even just `No Link` fits better.
Also the description to `Renders Logo/Image without any link`

engines/common/nucleus/particles/logo.yaml
```diff
@@ -20,6 +20,12 @@ form:
       type: input.imagepicker
       label: Image
       description: Select desired logo image.
+      
+    url_disable:
```
w00fz: Maybe `nolink` is a better naming?

marktaylor46:
Great stuff @adi8i But I agree with @w00fz that it should be "no link" for the field label.

w00fz:
Thanks @adi8i

mahagr:
Umm, there's one minor thing I don't like about this change: the option being `No Link`. I never like negative options as checking a checkbox really means enabling something, not disabling it. To me its counter intuitive to enable disabling something and I think its also considered as a bad practise by usability people.

As an afterthought it would have been better if the particle would have disabled link if you had empty link field. :) But I guess we cannot change that anymore, so this change is the second best thing to do.

marktaylor46:
@mahagr You can't do that because when the field it empty the particle uses the "home" page as the link instead i.e. there is never an option to have no link at all since blank for link actually means "home".

adi8i:
Maybe a dropdown would be better with the following options, link to: Home page, Particle Link, No Link.

With better naming ofc.

marktaylor46:
All I would say is you have to consider existing sites with "blank" url and those with link url completed - they must behave the same as they do now - I don't think a dropdown would give you that because you would have to default to one thing and which it was would be wrong in some cases.

mahagr:
@marktaylor46 That's why I said its an afterthought.

I'm mostly fine with the change, its good. The only thing I don't like is to have `No Link` instead of `Enable Link`.

mahagr:
Also it might be good to have a placeholder saying `Home Page` instead of just having empty text field, as with this new option it is a bit confusing what the meaning of empty have.

marktaylor46:
Yah I don't really mind how it gets changed as long as it's backwards compatible with existing sites. Also "home page" was my term it actually isn't that it's "default menu item" really (which is usually home but doesn't have to be).

adi8i:
https://github.com/adi8i/gantry5/commit/4f5740c14678ee0e991c942bafd96c2f750de683 @mahagr would be better like this ?

mahagr:
Yeah, and then default the option to 1 as well in yaml file.

PR #1622, opened 2016-10-17
Content Array: Add raw filter on title, for all platforms (wp, joomla, grav)

adi8i:


mahagr:
I'm against this change as Joomla for example assumes that titles are text only and need to be escaped or all kinds of bad things will happen. WP may have already escaped titles, making raw filter mandatory in there. @newkind ?

newkind:
Yah, WP passes title by couple functions so actually if you want to have HTML in the title you need to disable these actions/filters.

w00fz:
Going to close this as I also feel like this should not interfere with the platforms behavior.
Sorry!

PR #1638, opened 2016-10-24
Logo particle

adi8i:


mahagr:
:+1: I'll give others a change to check this change as well before committing it.

adi8i:
Of course, better approaches are welcomed.

w00fz:
Think we agreed this would stay as No Link but I don't care, either way is fine by me :)

PR #1917, opened 2017-03-09
Fix markup

Quy:


mahagr:
Thanks!

Quy:
@mahagr The quotes are escaped `data-g-assignments-parent=&quot;&quot;`. Would this be the right fix? `{{ group|raw }}`

mahagr:
Yes, I'll fix it.

mahagr:
Should be good now.

Issue #1887, opened 2017-02-22
Target option for Logo / Image particle


pdowse:
Hi Guys,

Would be great if there was a targeting option for the Logo / Image particle. For example if you use this particle to add a suppliers logo to a page. it makes sense that the image link opens in a new window. 

Having a target="_blank" or 'open in a new window' option at the particle level (instead of having to hard code it into the particle as an override) makes sense as then you can use this particle for a combination of internal and external links on multiple pages.

vmamet:
I went thru the same issue : 
my solution if it can help is to use custom html instead of logo/image
<a href="https://www.link.com" target="_blank">
<img border="0" alt="" src="images/test.jpg"> 
</a>
It worked for me ;)

PR #1918, opened 2017-03-09
Enhance Logo Particle

adi8i:
ref https://github.com/gantry/gantry5/issues/1887


adi8i:
I see, then I'll leave the `link` as it was, `input.checkbox` that why I used `true`, to keep it backward :).

PR #2045, opened 2017-07-06
Add a "global" rule for assignments per taxonomy

nikola3244:
In past, if user wanted to use a custom outline for entire taxonomy, they had to assign an outline to each term.
So that meant going back to outline assignments page each time when there's a new term, and adding new assignments.

But that user won't have to do that anymore, take this for exampe:
We have `Outline A` and `Outline B`, and of course `Base Outline`. We also have a taxonomy called `Tags`. In `Tags` taxonomy, there are `Travel` and `Cuisine` terms.

`Outline A` is assigned to `Tags - Archive View`.
`Outline B` is assigned to `Travel` tag.

If we open `Travel`, we will see `Outline B`.
If we open `Cuisine`, we will see `Outline A`.

If we remove `Outline A`'s assignment, and we open `Cusine`, we will see `Base Outline`.

This will simplify assignments for taxonomies and save user's time.


igor-gp:
@newkind any updates on this one? Thanks

webmaster-amka:
+1 for this. Any news?

Issue #2116, opened 2017-09-15
Error thrown when cloning section + PR


drnasin:
Hi guys, I noticed this bug few days ago but as I had a workaround I waited a bit to report.

The problem is following:

When using:
Section Settings->Inheritance
Mode: Clone
Outline: (any whose section to be cloned is empty and has no particles)

Replace: particles within sections - if checked causes error to be thrown (children is undefined in stdclass)

Problem is in this line:
https://github.com/gantry/gantry5/blob/develop/src/classes/Gantry/Admin/Controller/Html/Configurations/Layout.php#L457

Proposed fix:

if(is_object($item) && property_exists($item, 'children')) {
    $data->join('children', $item->children);
}

PR coming

PR #2117, opened 2017-09-15
Update Layout.php

drnasin:
https://github.com/gantry/gantry5/issues/2116

Add check if item is object and "children" property is defined.
Not sure if there should be "else" part...


mahagr:
I'm going to accept this change but I will simplify it as the layout can only be an object.

Issue #2166, opened 2017-11-06
Custom body classes / variants of the same outline


joomlabeat:
We need a way to be able to override the template files and also be able to add custom body classes like requested on these issues here:
https://github.com/gantry/gantry5/issues/949
https://github.com/gantry/gantry5/issues/2083

Actually, I am considering the most efficient way to create a color scheme variation based on some criteria.
The case is that I already have a few outlines, which work well - but now I would like to use those with a different coloring - if for example, the user is logged-in, or if an article is of a certain category and other similar criteria. I would prefer to not to have to create new outlines just to change the colors - but to use the same outlines with either an extra body class that will override the default colors, or by adding extra config options for outline's color scheme that will be used when those criteria exist - the latter seems impossible though.

What are the possible options?

joomlabeat:
Basically, I put it wrong - template files are overridable using the custom folder. What is missing is a way to manipulate overall what classes will be added to the body. So, we need a way to override/extend some of the framework's classes/methods.

joomlabeat:
Maybe you could give us a custom jinput array e.g. $gantryUserBodyClasses. We could add values into that array from a plugin and set it to jinput. Then in framework's page class, it can check for the $gantryUserBodyClasses array from $input and if it exists, it can loop over and add those classes to body. I just tested this and it works nice - without changing a lot in gantry.

mahagr:
Looks like this already works in Grav and WP.

In Grav you can have `body_classes` variable in header of the page.
WP uses `get_body_class()` function.
Joomla doesn't have native support for setting body classes, so it's missing this feature.

PR #2167, opened 2017-11-06
Joomla: Give G5 framework the ability to accept extra body classes

joomlabeat:
#2166 

This PR - opens up a window for additional customization by other Joomla devs. Gantry framework will check for a special reserved array variable in JInput - devs can use a custom plugin to assign elements on an array and then set that variable to JInput - If Gantry finds this variable, it will loop over the elements and assign those as additional classes to the document body.

This is just a minimum addition for those that are looking for this extra customization ability.


mahagr:
I'm not really happy with this kind of solution as IMHO it allows end users to customize the site without site owners approval.

joomlabeat:
@mahagr 
hey there... I agree with you on the first part that is: I am not really happy with this solution. My reasons though are different: 

It feels a bit "hacky" and it's not that elegant
It would require additional Joomla coding for someone to add custom classes to the body
it's not actual part of gantry that can be managed from the gantry admin (this last one sounds similar to what you said).

At the other side:
It's easy and simple implementation for gantry. 
It can be useful for template and other devs that need to overcome that small gantry's limitation, as it opens up that small window.  
It could be a small hidden feature inside gantry, till a more elegant or complete solution comes out. 

Having said that - I am still not sure what I decisions I would make, if I was the gantry framework developer.

*I already use it on one website with a custom gantry template. I will be ok to maintain this on my own after any Gantry updates.

mahagr:
Maybe an event to hook into instead of this, one that allows you to add body classes from a plugin. Would that solve your issues?

And no, I don't think I want to implement this in this way... :cry:

joomlabeat:
I started digging for a better way, but various life events distrusted me from anything work related. I am trying to catchup with my work at the moment. 
I will get back to this at a later time.

mahagr:
This pull request has no changed code. Closing.

PR #2214, opened 2018-01-17
[Bug] Copyright Particle Fixed

thexmanxyz:
I fixed the wrong namespace for link access within the `copyright.html.twig` because accessing the link information can not be achieved via `particle.owner.link`.  The correct namespace is actually `particle.link`.


N8Solutions:
@thexmanxyz Thank you for creating this pull request! I was just looking at the Copyright particle this past weekend and wondering why the link was not working.

thexmanxyz:
@N8Solutions np, I found it worth pushing because I noticed the bug as I used the Copyright Particle myself. Hope that the PR is merged soon because it is already open quite long. @mahagr @newkind

mahagr:
I'd prefer using empty target (`''`) if there's no target to be set.

thexmanxyz:
@mahagr First and foremost, thanks for merging. I'd prefer the empty target `''` to its sleeker. Just out of curiosity do we really need `particle.target|e`? Because we output the attributes `|raw` afterwards. So basically there is no need to escape at the point of constructing the attribute HTML.

mahagr:
You need to escape unsafe strings, meaning that if I generate HTML and use `|raw` filter afterwards.

thexmanxyz:
@mahagr Do we really need HTML context escaping for a parameter that is set to four predefined strings? I can't imagine myself a scenario where this might be unsafe at all...I don't want to bug you I simply want to understand the intention.

mahagr:
`|raw` should really not be used anywhere where there's (potentially) unfiltered content coming in, including any content in database and yaml files. The reason is that if you assume user input as safe, someone will find a way to eventually do XSS just because of the template may be used in an unexpected way, if not now, maybe in the future.

We're getting a lot of security reports for Grav Admin, just because of administrator can himself introduce XSS attacks. Yeah, we can ignore those in both Gantry and Grav, but they are still considered attacks. What if you allow someone limited access to the admin and he gets pissed off when he gets fired?

So basically all input which can be changed by any user, including administrator, needs to be escaped unless the content is specified as being HTML. And even in that case it should be filtered or reported if it has something potentially fishy in it.

thexmanxyz:
@mahagr Thanks for your explanations I really appreciate it. I'm fully aware of the basics of string escaping, XSS attacks and why you should never ever allow user generated content to be written out unfiltered. Except the case you laid down yourself when the user has elevated rights or the role which allows him to modify or add stuff intentionally without a malicious intent. 

But let me rephrase what I actually meant. We have in the same particle (ever had - even before my PR) fields that are written out `|raw` like [here](https://github.com/gantry/gantry5/blob/7735e5eac9f42526844362872f7e4a5bc7e072d7/engines/common/nucleus/particles/social.html.twig#L9). As I said this is an obvious case because we want the option for the `title` field to be entered as pure HTML. That's fine. But also every person who tinkers with the system or at least wants to do some damage and has the privileges to modify backend fields or the respective data files can simply exploit all these fields that are already there. And can even find them easily all over the place.

So I do understand what you are saying and I also understand that escaping more fields reduces the attack vectors however if someone wants to do damage this won't stop the person. Because as long as there are `|raw` outputs which obviously is the case on many places...escaping a fixed value field doesn't add any real security benefit. I accept your decision and your intention but I see no benefit when the same Particle has at least one field written out unfiltered as `|raw`. Even it reduced a potentially risk but not really helps much in the discussed case.

mahagr:
Personally, I consider all of those `|raw` filters as likely security bugs if the content is coming from a user. And yeah, we should be checking and fixing all of them. As I said, nothing coming from the user should be having `|raw` filter on it and there should be no exceptions on that. If you want the user to be able to pass HTML, there should be a new filter that checks the input against XSS.

thexmanxyz:
@mahagr You're totally right, the best way would be to completely eliminate all `|raw` outputs by applying an appropriate filter. I would like to say thanks for the very useful and constructive discussion on the topic. So basically there is already a specific filter just for HTML output planed? I assume this is not that trivial without cutting down functionality and still offering enough freedom. But TBH I don't know if there are already established methods which work good for that case (or even if symphony does offer in a newer version an predefined filter).

mahagr:
I guess that you cannot really make a safe HTML filter, but we have already started to look into this by detecting if a page is using certain HTML tags and displaying a message in admin if they are. In my example using `|raw` filter is the correct way to do (because of the HTML is being generated in the template file).

thexmanxyz:
@mahagr There are ways, one example is implemented by mail clients (e.g. Thunderbird) which only renders a reduced (light) set of HTML to prevent XSS and other malicious content that might be included in the mail body. However, it would obviously not be appropriate at all occurrences of `|raw` because it would definitely break things depending on where it is used.

mahagr:
Not only that, but stripping html is really, really slow in PHP.

PR #2241, opened 2018-02-15
Added redraw call to handle picture width problem in case scrollbar i…

Marmanen:
…s removed in Firefox.
Firefox shows part of the next picture as the right scroll-bar is removed when increasing the height of the window. This fix triggers redraw in this case and the picture width is increased to cover the area that was behind the scroll-bar.

### DIFF
--- a/themes/helium/common/js/owlcarousel.js
+++ b/themes/helium/common/js/owlcarousel.js
@@ -159,6 +159,8 @@
 			}
 		};
 
+        this._initialResized = false;
+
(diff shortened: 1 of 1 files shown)


mahagr:
@Marmanen You may want to fix this into the owlcarousel project.

@newkind Can you check the fix?

newkind:
Yah, I agree with @mahagr. This should be a pull request for OwlCarousel. If we add it here it will be reverted with any script update as no one will remember that it was modified before.

Issue #2273, opened 2018-04-10
mod_menu overrides in Joomla not working with Gantry Menu


N8Solutions:
I have discovered an issue with Firefox that is quite annoying and I have figured out a way around it using a Joomla! override.  However the override does not work for Gantry templates. I have applied the override to several different installations and the outcome is the same every time. It works on the standard Beez or Protostar template but when a Gantry template is applied it fails.
Here is a link to the discussion thread with the solution over on the Joomla! forum.
https://forum.joomla.org/viewtopic.php?f=706&t=960992

Any ideas on how I can get the override to work with the Gantry templates?
I believe it has something to do with the fact that the Gantry menu has it's own setting for the "Link Target" and the override is not passed through to it properly.

@w00fz I know the menu is your area of expertise though if you @mahagr or @newkind have any input for me I would welcome it very much!

N8Solutions:
Quick update, I found the file **Menu.php** with the following code from lines 310 to 323

```php
                // Get default target from Joomla.
                switch ($menuItem->browserNav)
                {
                    default:
                    case 0:
                        // Target window: Parent.
                        $target = '_self';
                        break;
                    case 1:
                    case 2:
                        // Target window: New with navigation.
                        $target = '_blank';
                        break;
                }
```

As a test I modified the **_blank** to **_self** but unfortunately the Gantry menu does not explicitly define and add **_self** to the link since it is the default. However, due to the behavior of a pinned tab in Firefox, unless it is explicitly defined the browser will open open it in a new Tab/Window.

I know this shouldn't be necessary but in this circumstance it is due to the way that a pinned Tab in Firefox is functioning.

thexmanxyz:
@N8Solutions How about overriding `menu.html.twig` shouldn't that be enough to create a custom fix?!

N8Solutions:
@thexmanxyz I'm not really sure if it would be or not. Like I said, I replaced **_blank** with **_self** in the `menu.php` file however because **_self** is the default it is not explicitly written. So even though I changed it I need to have a way for it to be defined explicitly otherwise it won't solve anything.

I had a look at the `menu.html.twig` file located in `\engines\common\nucleus\particles` and it imported `_self` as `SELF`. If there is a way to tell it to explicitly add `_self` to menu items that are listed as URL I do not know how to do it.

In the file there is this section

```twig
{% macro displayItem(item, menu, context, dropdown_type) %}
    {% import _self as SELF %}
[...]

PR #2276, opened 2018-04-12
Menu Item Link Target Extension - Firefox Tab Pinning Adaption

thexmanxyz:
Hey there, first of all I hope this Pull Request doesn't create a mess. If so I will cancel it and try something else but I already did a mistake as I didn't create a branch for my last PR which is still open.

So I had to create a branch from my fork based on the last commit before I commited to my fork to don't mess up commits in the second pull request. However from the preview it seems to work as expected ;).

Please take further information concerning this Pull Request from Issue #2273. I added another option to the menu item configuration which allows us to force the `target` attribute for menu links e.g. this is relevant for the target option `_self` were currently no `target` will be rendered to the DOM.

<del>But basically the changes add another option to the link target dropdown of menu items. This means the user can differentiate between:</del>

- <del>**none:** no target attribute will be rendered for the menu item link</del>
- <del>**self:** the target will be rendered as `target="_self"`</del>

This modification is not absolutely necessary but Firefox has a "bug" (it's not clear if it will be ever fixed) which breaks the target behaviour for pinned tabs if the target attribute hasn't been specified. Hence, FF forces tabs to be opened as they were attributed with `_blank` even thought there is no `target` attribute defined on that link. This modification gives users the option to shift around this issue by forcing `target="_self"` on menu links. The default behaviour is preserved => no target attribute is rendered.

Info: Please take also a look to the deleted stuff above. I reworked the PR because it's not possible to add another option to the Gantry menu target setting since it is connected with the Joomla menu target setting. Therefore I added the 'Force Target' checkbox instead.

### DIFF
--- a/engines/common/nucleus/particles/menu.html.twig
+++ b/engines/common/nucleus/particles/menu.html.twig
@@ -41,7 +41,7 @@
         {% set active = menu.isActive(item) ? ' active' %}
         {% set dropdown = item.level == 1 ? ' g-' ~ item.getDropdown() %}
         {% set parent = item.children ? ' g-parent' %}
-        {% set target = item.target != 'none' ? ' target="' ~ item.target|e ~ '"' %}
+        {% set target = (item.target != '_self' or item.target_force) ? ' target="' ~ item.target|e ~ '"' %}
         {% set rel = item.rel ? ' rel="' ~ item.rel|e ~ '"' %}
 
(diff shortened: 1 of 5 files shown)


mahagr:
@thexmanxyz What if we move this fix into the menu parameters instead of forcing people to enable this option for every single menu item?

mahagr:
Fixed in ab7081d

thexmanxyz:
@mahagr Thanks for fixing the issue and mentioning. I think the menu parameter option is the better fix!

thexmanxyz:
@mahagr Please also check my comment on https://github.com/gantry/gantry5/commit/ab7081d0e77bcd45e21295d6d5c584fcc86df5db. Don't want to open another PR just for this tiny change.

PR #2286, opened 2018-04-27
Owlcarousel Particle - Disable Slides 

thexmanxyz:
Hey there,

I decided to adapt the Owlcarousel Particle to support the functionality to disable single slides. I needed this functionality for my sites so I decided to push these changes to the Gantry repo. It's often useful to enable/disable single slides depending on special events e.g. christmas or easter. Hence, I think this is also useful for other people ;).


N8Solutions:
That's a great idea! This way you can keep the slides and use them again in the future without having to delete them and recreate them!

PR #2311, opened 2018-06-03
Update Google font list

paulxb:
I needed a font that was in Google fonts but not available in Gantry, so I updated the Google Font json file to the latest from Google.


mahagr:
@w00fz Can you take a look if this is good to merge?

newkind:
Looks good, I'll test it after merging. Thanks :)

PR #2348, opened 2018-08-21
hide submenu & indicator if no children are visible

Chrissi2812:
I don't know if this affects any Platform other than Joomla. I just have tested Joomla.
## Current Behavior
unfixed
Parent Items have the dropdown-indicator even when all children are unpublished or hidden.

## After this Fix
fixed
The Parent Indicator is only visible if there is at least one visible and published children.

*P.s The fix is also just tested with Joomla.*

### DIFF
--- a/engines/common/nucleus/particles/menu.html.twig
+++ b/engines/common/nucleus/particles/menu.html.twig
@@ -44,6 +44,13 @@
         {% set target = item.target != '_self' ? ' target="' ~ item.target|e ~ '"' %}
         {% set rel = item.rel ? ' rel="' ~ item.rel|e ~ '"' %}
 
+        {% set visibleChilds = false %}
+        {% for child in item.children %}
+            {% if child.visible and child.enabled and enabled|default(1) %}
+                {% set visibleChilds = true %}
+            {% endif %}
+        {% endfor %}
+
         <li class="g-menu-item g-menu-item-type-{{ item.type }} g-menu-item-{{ item.id }}{% if not item.dropdown_hide %}{{ parent }}{% endif %}{{ active }}{{ dropdown }} {% if item.url and item.children %}{% if not item.dropdown_hide %}g-menu-item-link-parent{% endif %}{% endif %} {{ item.class|default('') }}"
                 {{- SELF.getCustomWidth(item, menu, 'item', dropdown) }}
                 {%- if context.particle.renderTitles|default(0) %} title="{{ item.title }}"{% endif %}>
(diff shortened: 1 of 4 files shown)


simmonsr:
Thank you for your suggestion, perhaps @mahagr can consider it, but in the mean time you can disable indicator via Menu item settings (Disable Dropdown):

screen shot 2018-08-21 at 10 48 39 am

Chrissi2812:
@simmonsr I have just created an override for the menu particle, so I have a working solution in the meantime.

engines/common/nucleus/particles/menu.html.twig
```diff
@@ -44,6 +44,13 @@
         {% set target = item.target != '_self' ? ' target="' ~ item.target|e ~ '"' %}
         {% set rel = item.rel ? ' rel="' ~ item.rel|e ~ '"' %}
 
+        {% set visibleChilds = false %}
+        {% for child in item.children %}
+            {% if child.visible and child.enabled and enabled|default(1) %}
+                {% set visibleChilds = true %}
```
Chrissi2812: Should we break the loop?
Because if one child is visible we can skip the rest of the children.

I don't know if this is worth the effort?
Because then this TokenParser needs to be implemented... https://stackoverflow.com/a/40949346/6370595

mahagr: It doesn't matter if there are just a few children, but if the menu is huge, there will be some performance improvements on this. Generally, menus aren't that large, though...

newkind:
This definitelly needs to be tested across all platforms.

Chrissi2812:
I have tested it manually with Grav (1.5.3) and Hydrogen (v5.4.26) and the fix doesn't break the menu.
But in Grav it also works without this PR because it handles invisible menu items differently.

Or can the menu be filtered before it's handled by the twig template?
Maybe in [src/platforms/joomla/classes/Gantry/Framework/Menu.php](https://github.com/gantry/gantry5/blob/master/src/platforms/joomla/classes/Gantry/Framework/Menu.php#L265)?

Chrissi2812:
I've created a new branch where the menuItems are filtered before passing them to Twig. 

See here: https://github.com/Chrissi2812/gantry5/commit/e4e61367dc6fc51b7de8caff3cb461769980d444.

But this has the side effect that hidden menu items won't be shown in the Gantry Menu-Editor, but that's in Grav the same.

*PS. Should I open a new PR from the new branch then?*

mahagr:
@Chrissi2812 You can also update this pull request with merge.

PR #2366, opened 2018-09-17
Fix Content Tabs Particle Duplicate ID

thexmanxyz:
Hey I already noticed a while ago that there is an issue with the *Content Tabs* Particle. If you add it muliple times to a page it would create duplicate IDs for `g-contenttabs-item-`. Below I added a screenshot which shows the W3C results if I add Content Tabs multiple times. 

I decided to take the particle ID to build the item ids so we get unique IDs no matter how many particles we add. Probably it would be better to address everything with some other selector and completely drop the `id=` part and use `class=` instead. However I think it is a quick fix and obvious fix.

fix_content_tabs fw
.


thexmanxyz:
@np, thanks for the fast merge :)

PR #2396, opened 2018-10-20
fix(Filesystem): keep root link if excl. target

jguice:
When `$include_target` is false and the root `$folder` is a symlink, don't unlink/remove it.

e.g. when recompiling CSS via `ThemeTrait::updateCss` and the compiled css folder is a symlink itself, the current code immediately attempts to unlink `css-compiled` erroneously (resulting in a failure in some cases).  Instead, symlinks should only be unlinked on subsequent calls to `doDelete` (for which `$include_target` will default to true).


mahagr:
Good catch.

PR #2398, opened 2018-10-23
filter hidden menu items

Chrissi2812:
This is another approach to the issue I've tried to fix in #2348, but this will only affect Joomla.
It will change the menu editor behavior slightly, because hidden menu items won't show up anymore.
But that's the same for the editor in Grav.

### DIFF
--- a/src/platforms/joomla/classes/Gantry/Framework/Menu.php
+++ b/src/platforms/joomla/classes/Gantry/Framework/Menu.php
@@ -264,6 +264,11 @@ class Menu extends AbstractMenu
 
             $menuItems = $this->getItemsFromPlatform($params);
 
+            // Filter hidden items
+            $menuItems = array_filter($menuItems, function($item) {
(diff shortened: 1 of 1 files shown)


Chrissi2812:
Any news on this? It's been a year since I opened it.

Joomla's mod_menu is doing something similar in it's [mod_menu/helper.php](https://github.com/joomla/joomla-cms/blob/9c437d0de8035edc8903f8b138b89dfa658e2808/modules/mod_menu/helper.php#L75-L81) where this method original comes from. 

Maybe it's better to just port everything from there over?

mahagr:
The check should not be in this level, the same code is used in admin to edit pages.

Our code is delivered from some older version (J3.4 I think) of Joomla helper class and template file. I will need to take a look into this as I'm currently removing support for older Joomla versions.

monikat12titans:
I am closing this pull request because there has been no activity since 2019.

Issue #2399, opened 2018-10-23
No Logo Size Settings and No Instructions for how to resize it (Helium)


greenlanegreb:
It's a great shame, that, despite a plethora of options for particles and modules that there is no built in ability to either allow for the "real size" of a logo image to take precedence in terms of the size that an uploaded logo displays at or any meaningful settings to allow for setting the width and height. 

I've searched for straight forward instructions but there are none within the Gantry instructions. I'm reporting this as a bug as developing a website with a logo is a very basic requirement and, despite the ease with which other tasks can be accomplished, this has been well and truly overlooked and not even detailed in the instructions. 

I'd like to see size settings introduced into Gantry to allow for sizing and for some (decent) instructions to be added into the manuals. 

Thanks :)

mahagr:
Just remember to keep the particle simple enough and backwards compatible: it should keep on working as it was without needing to save/change the particle. :)

PR #2424, opened 2018-11-25
Add logo image max height

adiwebdev:
It works with logo images only, as svg size shouldn't be set using width / height. The default value is empty, it won't affect existing websites. I tested with Helium theme.

ref: https://github.com/gantry/gantry5/issues/2399


mahagr:
@newkind Can you check this out?

newkind:
Looks good to me. @adiwebdev can I ask you to modify the label in YAML from `Image Height` to `Maxium Height` or something like this? Using `Image Height` can be confusing when your image is actually shorter than the max height value.

PR #2441, opened 2018-12-30
Extension of owlcarousel particle - planned / timed slides

thexmanxyz:
@newkind @mahagr 

This PR can be seen as an addition to a previous merged PR (#2286) where I added the disable functionality for owlcarousel slides. I now had the time to create a more sophisticated solution for the Particle which I already had in mind for months.

It is fine to have slides that can be disabled on demand but even better for maintenance of a websites would be to have specific time intervals to show slides. However, that still wasn't enough, it should be possible to define recurring events (e.g. new year, x-mas, holidays, whatever) and not within a specific one-time frame.

Therefore I extended the previous behaviour so it is still possible to enable / disable items like before. Additionally, I added two improved options which allow us to show slides on a one-time or an annually event basis. To sum things up we have now the following options to show slides:

- **yes:** show always
- **no:** show never
- **annual:** show every year when the current date is in the time frame
- **once:** show only when the current date is in the exact time frame

### DIFF
--- a/themes/helium/common/particles/owlcarousel.html.twig
+++ b/themes/helium/common/particles/owlcarousel.html.twig
@@ -9,9 +9,15 @@
 
             {% for item in particle.items %}
                 {% set showItem = item.showItem|default('yes') %}
-                {% if showItem == 'once' or showItem == 'annual' %}
-                    {% set dFormat = (showItem == 'once') ? 'Y-m-d' : 'm-d' %}
-                    {% if item.startDate|date(dFormat) <= "now"|date(dFormat) and item.endDate|date(dFormat) >= "now"|date(dFormat) %}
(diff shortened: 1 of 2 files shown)


thexmanxyz:
@mahagr I probably need to remove `caching` [here](https://github.com/gantry/gantry5/blob/7fda10941609afac31f846169120d044f9735d72/themes/helium/common/particles/owlcarousel.yaml#L7-L8) to allow the feature to work correctly but I'm not sure. Provide me some input and I will add the respective commit removing it.

mahagr:
Just FYI on caching, the static caching is meant to use on a content that is known never to change (unless the particle gets saved).

thexmanxyz:
@mahagr Alright, so we need to remove caching here to avoid issues. In the meanwhile I also tested that with some events and confirm that. Will prepare a commit in the next minutes. Thanks for additional feedback.

PR #2443, opened 2019-01-03
Use Gantry classes instead of Bootstrap classes

uglyeoin:
This update uses the Gantry5 classes and grid instead of the Bootstrap one.  I don't know why Bootstrap was used in the first place, maybe there is a reason, in which case please feed back.


uglyeoin:
I see, ok then I will close.  Thanks for the response.

Issue #868, opened 2015-09-23
Block Size 66.6 


Chrissi2812:
I want to have the sidebar the same width as the blocks in the mainbottom section. 
![image](https://cloud.githubusercontent.com/assets/6141652/10049907/fe10c000-6219-11e5-8971-ec7dc9b3a7cf.png)

But there is no class 66-7 or 66-6, so i gave the content block a class with this name but then i can't save the block because of the 66.7% size.
And when i change it to 66 the sidebar changes it's size too.
![image](https://cloud.githubusercontent.com/assets/6141652/10050151/33cd4ea6-621b-11e5-9eed-cccec03f76a3.png)

it works when i first give the content section the class and a width of 66% or whatever and then give the sidebar a width of 33.3%. But that feels quite complicated.

adi8i:
Maybe Devs have other approach or a better idea, but for now, the easiest way i can think at, is to assign a class on `container`. Set mainbody to 67% and sidebar to 33% (you can chose any other values, but target them in css too).

![1](https://cloud.githubusercontent.com/assets/9744973/10052774/3aac8e2e-6231-11e5-926e-2eb19ce66206.PNG)

Then you can add the following code in `custom/scss/custom.scss`

```
@import "dependencies";

@include breakpoint(desktop-only) {
    .custom-size {
        .size-67 {
            flex: 66.7%;
            width: 66.7%;
        }
    
        .size-33 {
            flex: 33.3%;
            width: 33.3%;
        }
    }
}
```

w00fz:
We don't technically support decimals on sizes, the only cases where you get decimals is when a grid is equalized. So if you drop 3 particles in an empty row, you will get `33.3 | 33.3 | 33.3`, if you remove one, you will get `66.7 | 33.3` and so on.

The moment you start dragging to resize, you will loose the decimal, you can always equalize through the equalize button that you get on the far right when hovering a grid:

![equalizer](https://cloud.githubusercontent.com/assets/11734/10056380/b26d5a02-61ef-11e5-8e44-12a70204a90a.jpg)

With all that said, what you want to do is to simply start dragging the blocks so they loose the decimals, making it easier to control the same size aspect, and you will end up with something like this:

![same](https://cloud.githubusercontent.com/assets/11734/10056472/3846e076-61f0-11e5-802a-6e8c08a71e41.jpg)

Chrissi2812:
I think i will go with the approach of adi8i, it really seems the easiest way.

[...]

PR #2455, opened 2019-01-30
Update _sizes.scss

uglyeoin:
Relates to #868

Nucleus currently has these sizes:

```
.size-33-3 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 33.3333333333%; -moz-flex: 0 33.3333333333%; -ms-flex: 0 33.3333333333%; flex: 0 33.3333333333%; width: 33.3333333333%; max-width: 33.3333333333%; }
.size-16-7 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 16.6666666667%; -moz-flex: 0 16.6666666667%; -ms-flex: 0 16.6666666667%; flex: 0 16.6666666667%; width: 16.6666666667%; max-width: 16.6666666667%; }
.size-14-3 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 14.2857142857%; -moz-flex: 0 14.2857142857%; -ms-flex: 0 14.2857142857%; flex: 0 14.2857142857%; width: 14.2857142857%; max-width: 14.2857142857%; }
.size-12-5 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 12.5%; -moz-flex: 0 12.5%; -ms-flex: 0 12.5%; flex: 0 12.5%; width: 12.5%; max-width: 12.5%; }
.size-11-1 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 11.1111111111%; -moz-flex: 0 11.1111111111%; -ms-flex: 0 11.1111111111%; flex: 0 11.1111111111%; width: 11.1111111111%; max-width: 11.1111111111%; }
.size-9-1 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 9.0909090909%; -moz-flex: 0 9.0909090909%; -ms-flex: 0 9.0909090909%; flex: 0 9.0909090909%; width: 9.0909090909%; max-width: 9.0909090909%; }
.size-8-3 { -webkit-box-flex: 0; -moz-box-flex: 0; box-flex: 0; -webkit-flex: 0 8.3333333333%; -moz-flex: 0 8.3333333333%; -ms-flex: 0 8.3333333333%; flex: 0 8.3333333333%; width: 8.3333333333%; max-width: 8.3333333333%; }
```
But 12 column grids have more sizes. To supplement the 7 sizes available we should add another 5.

100% / 12 columns = 1 column. (8.333333333333333)
1 column * X = X Columns.

(100/12) *5 = 41.66666666666666%
(100/12) * 7 = 58.33333333333333%
(100/12) * 8 = 66.66666666666667%
(100/12) * 10 = 83.33333333333333%
(100/12) * 11 = 91.66666666666667%

This pull request adds those sizes


uglyeoin:
@mahagr I think I have done this correctly, however, you might want to take a look over it to make sure (a) I've edited the right file and (b) my logic makes sense.  

I'm not sure why the other divisions were there but they all seemed to do it by division instead of multiplication.  The way to do that would be 1100 / 12, 1000/12, 800/12, 700/12, 500/12

newkind:
Ok this needs to be talked through - we need to check if this is not something related to the layout manager and how it generates classes etc.

adiwebdev:
This is related to the way `layout manager` works, adding those classes wouldn't be a big deal, since you won't be able to make use of them. It might be too late to add this classes anyway, changing the way columns are resized might have unexpected results on existing websites.

uglyeoin:
What I want to achieve is the possibility of converting a 12 column layout e.g. Bootstrap or one of the many, many others that use 12 column layouts, to a Gantry template easily.  So all I really want to do is to add the `66-6` size to the css, which I can do with an override without an issue.  It's only because @mahagr mentioned in the other issue that I should do a pull request.  Happy to discuss, I definitely don't want to break anyone's sites or the layout manager.  

To be frank... I mainly wanted to do this because compents/modules in Joomla! are usually based on Bootstrap, like Joomla!  But I hate the inconsistency so I want to convert them via an override to Gantry `g-grid`, `g-block`, `size-x`.  And the buttons from `btn` to `button`.  I can do this in custom.scss if this is a problem doing it in the css of Gantry.

newkind:
@uglyeoin this is something that can potentially cause conflicts with themes and layout manager, that's why we have to be very careful. What I'd suggest you currently is to simply add these classes to your theme or Nucleus overrides inside of your theme. That way your theme will override what was set in the nucleus.

uglyeoin:
Sure I can just add a new _bootstrap.scss and then call it in custom.  That way it's easy enough to reuse.  I'll close this issue.

PR #2509, opened 2019-06-25
Add option to disable particle items

thexmanxyz:
I added an additional field to all relevant particles which contain a `collection.list` to allow that a user can selectively disable certain front-end items. The idea for this adaption comes from @yellowwebmonkey. The particles which I found and come into question are:

* Social
* Sample (Hydrogen)
* Content Cubes
* Horizontal Menu
* Content Tabs

Please also see the PR #2286 which contains the same functionality for the Owl Carousel.

### DIFF
--- a/engines/common/nucleus/particles/social.html.twig
+++ b/engines/common/nucleus/particles/social.html.twig
@@ -9,12 +9,14 @@
     {% if particle.title %}<h2 class="g-title">{{ particle.title|raw }}</h2>{% endif %}
     <div class="g-social {{ particle.css.class }}">
         {% for item in particle.items %}
-            {% set title = (item.title is not empty) ? item.title|e : item.text|e %}
-            {% set titleAttrib = (title is not empty) ? ' title="' ~ title ~ '" aria-label="' ~ title ~ '"' : '' %}
-            <a href="{{ item.link|e }}"{{ targetAttrib|raw }}{{ titleAttrib|raw }}>
-                {% if particle.display in ['both', 'icons_only'] %}<span class="{{ item.icon|e }}"></span>{% endif %}
-                {% if particle.display in ['both', 'text_only'] %}<span class="g-social-text">{{ item.text|e }}</span>{% endif %}
-            </a>
+            {% if not item.disabled %}
+                {% set title = (item.title is not empty) ? item.title|e : item.text|e %}
+                {% set titleAttrib = (title is not empty) ? ' title="' ~ title ~ '" aria-label="' ~ title ~ '"' : '' %}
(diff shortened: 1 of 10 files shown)


thexmanxyz:
Tagging people @JoomFX @w00fz @n8solutions @yellowwebmonkey

JoomFX:
@thexmanxyz This is a great feature but in my opinion it should be implemented on the framework level somehow.
I'm simply not interested in modifying 70+ particles to add this feature :)

thexmanxyz:
@JoomFX Yeah that is also what I pointed out ;). A ideal solution would apply this automatically and handle the filtering of disabled ones.

mahagr:
@thexmanxyz I just have a single major wish: Please change the option to be positive instead of negative. This is for usability as it is better to keep enabled (green) state on left and disabled (red) on the right. :)

thexmanxyz:
@mahagr  I'm not sure what you mean because

```
.disabled:
  type: input.checkbox
  **label: Disabled**
  description: Enables or disables this item.
  **default: false**
```

The original commit was the other way around **Enabled Label** / `default: true` but obviously that won't work for `collection.list` items because they would not reconfigure automatically to default state. What that means is, that already existing items would be disabled until each list item is saved again. That's the reason I choose it this way. It's the same as in #2286 

Also there is not red and green because it is a default checkbox.

thexmanxyz:
@mahagr I updated the particles. I changed the description, the type is now `enable.enable` as you requested. I also changed the label to Enabled and on by default. However the issue from above is still there as new controls don't hold the default state until re-saved. But I now simply forced `|default(true)` on the control value which solves it for me. Now everything should be as requested.

mahagr:
Yes, you need to enter a default value in order to make it work before saving. That is fully expected as otherwise the default is null (not set).

thexmanxyz:
@mahagr I'm not sure if we are talking about the same. If I set up the following control:

```
        .enabled:
          type: enable.enable
          label: Enabled
          description: Enables the item on the front end.
          default: true
```
in a `collection.list`. The default state for `.enabled` (`true`) is only set if I open each already existing item and save it. Otherwise the state would be null (not set). That is what I experienced. Just reiterating the issue.

mahagr:
Yes, we are. What you're talking about is the default behaviour for configuration. If the value is not set, Gantry thinks it is null even if you had the default set in the blueprints. Blueprints are only used for displaying the form and saving it.

thexmanxyz:
@mahagr Alright, I'm glad that we talking about the same thing. Ok I now understand how the default values work! Thanks.

Issue #2536, opened 2019-09-07
Main menu visible on mobile devices during page load


thexmanxyz:
Hey there,

IDK if this one has already been reported but after @jozwkat pointed out in the Gitter chat room I also remembered that I faced the same issue on my sites and actually never reported it. But I will now finally state the issue.

On mobile devices (e.g. hamburger menu is active and replacing main menu) there might be a short period of time during page load when the actual desktop main menu is rendered before it disappears. If I am remembering right I had that issue on Hydrogen, Helium and AFAIK also on Photon. I'm only certain that it happened on Chrome Android but I think it also happens on desktop browsers when enable device testing or shrinking window. It might also has something todo with rendering speed of the machine. 

I also have fixed the issue with the following CSS rules:

```CSS
@media only all and (max-width:47.938rem) {
    .g-main-nav {
        display: none;
    }
    .g-block-mm  {
        display: none;
    }
}
```

This might not be the perfect solution but it gives an approach of solving the problem. `.g-block-mm` might not be necessary but I included it anyway. See the screenshot where this custom class was added.

![grafik](https://user-images.githubusercontent.com/17965908/64472878-4faab380-d165-11e9-9a11-b5c5195e5c2c.png)

jozwkat:
Great work! The problem is gone.

thexmanxyz:
@jozwkat Glad to hear that :-)

N8Solutions:
@thexmanxyz This is something I remember being discussed quite a long time ago, Feb. 2016! It goes to show how long I've been around the Gantry 5 project, LOL!
There was a lengthy discussion about this that included @hennysmafter & @JoomFX.
Unfortunately I'm having a problem going back very far with Gitter but I was able to find a message that I believe to be the first reply from @hennysmafter to @HenriKorhola to the start of the conversation.
:two: [February 16, 2016 10:41 PM](https://gitter.im/gantry/gantry5?at=56c3896efdaaf5f17c0b252c)
@JoomFX then replied here :three: [February 17, 2016 12:57 AM](https://gitter.im/gantry/gantry5?at=56c3a93fa5bdce025f69ba3e)
and the conversation continued from there but I'm not sure we did anything other than adding a code similar to what you have suggested here.

However, it was @JoomFX that first brought this to the attention of @w00fz a month prior to that conversation here :one: [January 8, 2016 12:14 AM](https://gitter.im/gantry/gantry5?at=568ee33fee13050b38a246f2)

[...]

PR #2538, opened 2019-09-10
Disable menu with CSS on mobile devices

thexmanxyz:
This PR is based on the issue #2536. It adds to the menu particle a checkbox option to completely disable the menu on mobile devices (`mobile-only`). For further information read the referenced issue which explains why this might be useful for users (flickering).

Hence, this gives the user the option to decide whether he wants to risk ending with no menu on mobile devices when no JS is supported but prevents the main menu to be rendered during page load (which is still an annoying issue for many people). This might be a good compromise because the default behaviour is maintained and there is an option to decide whether the risk should be taken (also explained in the field description).

Note: The SCSS added might not be added on the right spot and / or applied only for Hydrogen. Any input would be helpful I will change the PR properly if I know to do it correctly.

@n8solutions

### DIFF
--- a/engines/common/nucleus/particles/menu.html.twig
+++ b/engines/common/nucleus/particles/menu.html.twig
@@ -123,7 +123,7 @@
 {% import _self as macro %}
 
 {% if menu.root.count() %}
-<nav class="g-main-nav" role="navigation"{{ particle.mobileTarget ? ' data-g-mobile-target' : '' }} data-g-hover-expand="{{ particle.hoverExpand|default('true') ? 'true': 'false' }}">
+<nav class="g-main-nav{{ particle.disableMobile ? ' g-hide-mobile' : '' }}" role="navigation"{{ particle.mobileTarget ? ' data-g-mobile-target' : '' }} data-g-hover-expand="{{ particle.hoverExpand|default('true') ? 'true': 'false' }}">
     <ul class="g-toplevel">
         {{ macro.displayItems(menu.root, menu, _context) }}
     </ul>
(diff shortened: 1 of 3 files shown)


thexmanxyz:
@mahagr Even thought this PR was already merged into develop I have to mention again that the CSS rules applied within the `_nav.scss` file in the commit b760419 might only affect the Hydrogen template. I'm pretty sure that the changes made to `_nav.scss` are not used by Helium template.

So the rule has to be applied at the right position for the Helium template or the rules have to be moved to a SCSS file which is used by both templates.

thexmanxyz:
Adding also the reference to ebbc0b4 because I overlooked the `menu.yaml` file located under `engines/common/nucleus/particles`.

mahagr:
YAML change was already added by me in 2c68fc8.

@newkind I think the CSS change could be moved to the core nucleus?

N8Solutions:
@thexmanxyz & @mahagr Did this ever get implemented? I don't see an option for it in the menu particle for either Joomla or Wordpress.

mahagr:
I did add the option to the other platforms, hmm.

N8Solutions:
@mahagr is it possible I'm not looking in the right place? I just dont' see it in the menu particle.

thexmanxyz:
@N8Solutions You are not looking on the right spot actually. It was implemented and is available on Joomla and should also be there on WP. It is not available within the menu particle but rather under the Gantry 5 backend under `Styles`. 

grafik

Take a look here:

grafik

thexmanxyz:
@N8Solutions Please also check #2574 that question was already asked by @hacki65.

thexmanxyz:
@N8Solutions Additional note, it is only available for Hydrogen and Helium.

Issue #2196, opened 2017-12-13
Add support for Fontawesome 5


marktaylor46:
Since Fontawesome 5 is now available and uses SVG's it would be nice to provide support for this within Gantry5. Fortunately Fontawesome do provide a mechanism to retain backwards compatibility with fontawesome 4 codes through the use of their "shim" - please see...

https://fontawesome.com/how-to-use/upgrading-from-4

pplopx:
+1

Thanks!

810:
any progress?

wbarnes:
+1 Thanks,
 I am also interested in updates?

XaRz:
+1 Thanks!
we want new fa5! 👍

adoewa:
+1 tnx

webhelpus:
I'd like to be able to 1) easily use the new Font Awesome 5 icons (which are amazing) and 2) be able to use the Pro version too. (they just released their beta CDN)

pplopx:
Indeed, the CDN Pro of FA5 is now online (beta)! 

A suggestion: to implment the free version it is enough to include the files (and the corresponding code) in the distribution of Gantry, as usual. I understand that the solution for FA5 Pro users should be to be able to use this new CDN. I have not yet studied the method well, but it seems that each user would have to include a token that FA provides us with. 

Since I suppose that the distribution of Gantry can not include the exclusive icons of the FA5 Pro subscription by license terms, I suggest that the Gantry implementation offer as an alternative a setting that if it is filled with the code of the token, ignore the files of the version free included with Gantry and work directly with the CDN Pro. 

I think it should be possible to do so and it could be a smart solution. I can provide Gantry developers with access to the test token if they do not get it directly from the FA guys.

webhelpus:
For Now, pro users can add the CDN to their Head in Page Settings.  You'll have to enter the codes and won't have a GUI in Gantry.

paolavness:
Is there an ETA on this? Specifically, it would be good to be able to access the Icons in `simplecontent` and `infolist` particles in Grav. 

Could you give guidance on where to edit these particles in the backend to access the new icons?

thexmanxyz:
@webhelpus better way would be to override `page_head.html.twig`, `base.html.twig` and `layout.html.twig` this way it would be general available however IDK if there are any incompatibilities. Didn't do a test.

thexmanxyz:
[...]

PR #2540, opened 2019-09-12
Added Font Awesome page setting with disable switch

thexmanxyz:
The discussion on Font Awesome 4 and the successor Font Awesome 5 is already open for quite some time (#2196). As a first step towards supporting v5 I would recommend adding at least a switch to disable the front-end loading of the Font Awesome 4 library. This allows users to easier switch versions on the front-end without the need to override core files or double loading both versions. I decided to create a new panel within the Page Settings. It might not be the best place to handle that but probably more options would be necessary in the future concerning icons or especially Font Awesome so a separate panel sounds useful to me.

grafik

I added a screenshot to show what the PR does. This also would make things easier when writing an Atom which supports FA5. Because no need for me as a developer to touch core files. I can completely rely on that setting + an Atom which handles everything else. This also gives the community the power to handle FA5 without the necessity to wait on back-end support.

@mahagr @marktaylor46 @newkind @yellowwebmonkey @n8solutions

### DIFF
--- a/engines/common/nucleus/blueprints/page/fontawesome.yaml
+++ b/engines/common/nucleus/blueprints/page/fontawesome.yaml
@@ -1,0 +1,12 @@
+name: Font Awesome Settings
+description: Configuration for Font Awesome icon set and toolkit. 
+type: global
+
+form:
+  fields:
+  
+    enable:
+      type: enable.enable
+      label: Enable
+      description: Enable or disable the loading of the Font Awesome icon library on the frontend. This is useful if you want to manually add a different version of the library (e.g. v5.x).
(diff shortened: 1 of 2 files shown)


thexmanxyz:
@mahagr Thanks for merging the PR!

N8Solutions:
I just looked over this pull request. It's quite a very simple, yet elegant solution! Thank you for creating it @thexmanxyz !

thexmanxyz:
@N8Solutions You are very welcome! I hope this useful for other people too. I'm now looking forward on creating the anticipated FA5 Atom.

Issue #2534, opened 2019-09-06
Question | TWIG: Joomla Finder and Tags 


thexmanxyz:
@yellowwebmonkey @mahagr @newkind 

Is there a way to filter Joomla articles with the `joomla.finder()` by tag or alternatively return the tags with the article object? By default the objects returned do not contain the tags as array / attribute, checked with `dump(...)`. However there is a field `["show_tags"]=> string(0) ""`. I tried messing with that field as an option parameter but that didn't work. Some hint would be highly appreciated. Thanks!

I added a screenshot to show what field within a Joomla article I mean / target.

![grafik](https://user-images.githubusercontent.com/17965908/64416233-1a379480-d097-11e9-834e-e53ad7063a7e.png)

mahagr:
Unfortunately, tags are not supported by the finder.

thexmanxyz:
@mahagr @yellowwebmonkey Are there any plans to implement this feature within the finder or possibility to see this coming at some point in the future? Because there is a demand to fetch articles by their tags.

It would be great to at least have the option to fetch them by a method like:
`finder.tags(['tag-a','tab-b','tag-c'], match_all = true)`.

`.tags([tag_array], 'match_all')` ... it would be useful to have a second parameter which defines whether all tags (`true`) should be matched or any of the passed tags (`false`).

mahagr:
There is a feature request for this: #1631

I would love to implement this soon, but for me getting Joomla 4 support and menu fixes done are more important at this moment.

If someone likes to look into this (mostly if there's a way to use Joomla tags by some API), I would love to hear about it.

thexmanxyz:
@mahagr I completely missed the #1631 feature request. In the meanwhile we have a working PR for this FR so I will close this issue because it is not needed any more and everything discussed / outlined here should be working with #2561.

Issue #1631, opened 2016-10-21
Content Array particle should allow to select single tags


JoomFX:
Hey guys,

This issue is kind of related to https://github.com/gantry/gantry5/issues/1541

In order to have a really flexible and powerful particles that fetch platform content the user should be able to select single/multiple Tags for the "Data Source". Therefore we need new YAML fields (`joomla.tags`, `wordpress.tags` and probably `grav.tags`).

Probably the logic for WP is already there, since WP provides an API for those things, but I'm not sure if the logic is in place for the custom API that Matias implemented for Joomla.

As a side note, this is a great example where the "Dependent Fields" feature can (and should) be used (I know that it will not come until the Admin re-write, I'm just giving an example 😉 ).
Let's say that there is a "Data Source" `select.select` field in the "Content Array" particle. This field will have 3 options - "Categories", "Articles" and "Tags". Depending on the option the user selects, another field will appear:

If "Categories" is selected - `platform.categories` field will appear
If "Articles" is selected - `platform.articles` field will appear
If "Tags" is selected - `platform.tags` field will appear

mahagr:
I have not yet implemented tags support for Joomla Articles, which means that this feature was left out intentionally. I wanted to replicate most of the behaviour from Joomla content modules and they were lacking tags support.

As I do want to add support for tags (its already kind of supported by both Grav and WP), I'll assign this to Future releases.

JoomFX:
Thanks Matias.

adi8i:
As a note, in WordPress you can filter by tags using `wordpress.categories` field.

BjoernH01:
I would also like to see fields for tags being implemented.
That way we could get rid of RokSprocket and use particles for fetching and showing articles.
Since RokSprocket will probably not get adopted to Joomla 4, please make particles ready to substitute it.

BrentCritch:
Here's another vote for tags in Joomla. I'm working on a site that needs Content Pro to pull articles based on various criteria. I figured I could do it with a MultiCategories setup (http://www.cwjoomla.com/extensions/cw-multicategories) but they store the category data in attribs rather than catid so Content Pro won't pick it up. Tags would give me another way to accomplish this - at the moment I'm a bit stuck.

StefanvDelft:
Hi all,

I would like to add an extra post to let know that adding this feature would be great and very much appreciated! Thanks for all the good work :)

With kind regards,
Stefan

biltongandpotroast:
[...]

PR #2561, opened 2019-10-31
Added Tags filter for Joomla! articles

Kubik-Rubik:
This PR solves issues like #2534, #1631 and other.

**Usage in the content-pro-joomla particle** (as an example, use accordingly in other particles)

_YAML file_

```
article.filter.tags:
    type: input.text
    label: Tags
    description: Enter tag IDs to limit the articles that should be shown. It should be a list of tag IDs separated with a comma (i.e. 1,2,3,4,5).
    overridable: false
```

_Twig file_ (add it before the find() line(!))

```
{# Tags #}
{% set tags_options = filter.tags ? {id: [filter.tags|replace(' ', '')|split(',')]} : {} %}
{% do article_finder.tags(tags_options) %}
```

Demo video: https://downloads.kubik-rubik.de/joomla/gantry5-tags-filter.mov

### DIFF
--- a/src/platforms/joomla/classes/Gantry/Joomla/Content/ContentFinder.php
+++ b/src/platforms/joomla/classes/Gantry/Joomla/Content/ContentFinder.php
@@ -148,4 +148,16 @@ class ContentFinder extends Finder
             }
         }
     }
+
+    public function tags($tagIds = [])
+    {
+        if (empty($tagIds['id'][0])) {
+            return $this;
+        }
+
+        $tagIdsArray = $tagIds['id'][0];
+        $this->query->join('INNER', '#__contentitem_tag_map AS t ON t.content_item_id  = a.id');
+
+        return $this->where('t.tag_id', 'IN', $tagIdsArray);
+    }
 }


mahagr:
@Kubik-Rubik Thanks! I've been busy on other things and thus I've not had time to figure out how to integrate tags to the query. I do remember that tags were quite complicated as I did fix some issues with them years ago. 

Do you have reference to the code which does the same in Joomla module? I want to be sure there isn't more into it which could cause unexpected behavior when using JOIN to tags table.

thexmanxyz:
@mahagr @Kubik-Rubik @yellowwebmonkey I verified the PR right now and it does work but TBH I don't like that the feature was implemented only based on the Tag ID which is quite difficult to handle for end users. Moreover like suggested in #2534 I would love to see the option to match `ALL` or `ANY` depending on a parameter. Nonetheless I will now add a commit to this PR which adds the functionality:

1.) to filter for both (id and name)
2.) to match for `ANY` and `ALL` like suggested in #2534 

Here is the PHP code to match for tag name instead of ID (just for the sake of completeness):

```PHP
public function tags($tagIds = [])
{
    if (empty($tagNames['name'][0])) {
        return $this;
    }
    $tagIdsArray = array_map('intval', $tagNames['name'][0]);
    $this->query->join('INNER', '#__contentitem_tag_map AS t ON t.content_item_id = a.id');
    $this->query->join('INNER', '#__tags AS ta ON t.tag_id = ta.id');
    return $this->where('ta.title', 'IN', $tagIdsArray)->where('t.type_alias', '=', 'com_content.article');
}
```

thexmanxyz:
> 
> 
> @Kubik-Rubik Thanks! I've been busy on other things and thus I've not had time to figure out how to integrate tags to the query. I do remember that tags were quite complicated as I did fix some issues with them years ago.
> 
> Do you have reference to the code which does the same in Joomla module? I want to be sure there isn't more into it which could cause unexpected behavior when using JOIN to tags table.

What unexpected behaviour do you have in mind? Actually there shouldn't be any unexpected side-effects because the tags are filtered on `type_alias` and joined correctly over the intended FK.

thexmanxyz:
> 
> 
> @Kubik-Rubik Thanks! I've been busy on other things and thus I've not had time to figure out how to integrate tags to the query. I do remember that tags were quite complicated as I did fix some issues with them years ago.
> 
> Do you have reference to the code which does the same in Joomla module? I want to be sure there isn't more into it which could cause unexpected behavior when using JOIN to tags table.

@mahagr What unexpected behaviour do you have in mind? Actually there shouldn't be any unexpected side-effects because the tags are filtered on `type_alias` and joined correctly over the intended FK.

Kubik-Rubik:
Sorry for the late reply, was too busy! 

I prefer the IDs more but we could add a second function for the Tag names to have both options.

@thexmanxyz Or do you want to do the commit that combines it all?

thexmanxyz:
@Kubik-Rubik Don't worry! You were actually quite fast answering (~1 hour). I'm already in testing phase. I would not add a second method I will propose an update for 1.) in the next minutes. Stay tuned. I prefer to have everything within one method. Just define e.g.

```
{id: [ids], name: [names]}
```

But don't bother it is already implemented and working. I will add my updated version of the `tags()` method to this PR.

mahagr:
I'm currently working on Gantry 5.5, this week I've been cleaning up the code for PHP 5.6/7.1. So if you can work on this one I would appreciate. :)

Can you check the code against Joomla code, to make sure we're using tags in the same way as they are?

I agree that we need support for working with both tag ids and names.

thexmanxyz:
@Kubik-Rubik Can you please give me access to you fork? So that I can commit my changes directly?

thexmanxyz:
@mahagr @Kubik-Rubik @yellowwebmonkey @n8solutions I'm finally done with the rework. The signature of the method now looks like this (pseudo-code):

`tags({id: [ids], title: [titles]}, matchAll = false)`

- `id`: specify a list of tag ids
- `title`: specify a list of tag titles
- `matchAll`: match for all titles or ids

1.) First and foremost I added the functionality to filter articles for both tag `id` and `title`. As already outlined in my previous comment. You can specify both of them and they are matched based on a SQL `OR`. So if a tag `id` is found for an article it will be returned as well as when the `title` matches.

2.) Additionally I added the functionality to allow users to execute a full match against tag `ids` as well as tag `titles`. That means, only articles are returned which full match a list of tag `titles` or tag `ids`. 

**To clarify how this works:** 

If I have an article with the added tags `tag1`, `tag2` and `tag3` and I specify something like this: `tags({title: ['tag1', 'tag2']}, true)`. The article will be returned.

If I call the `tags(...)` method again with `tags({title: ['tag1', 'tag4']}, true)`. The article will not be returned because not all specified tags match the tags contained by the article.

This does work interchangeably with both tag `titles` and `ids`. So if the `titles` don't match but the `ids`  do match the article will be returned. This is also the same when omitting the `matchAll` parameter or setting `matchAll = false`.

**Additional**

I had to update the `Finder.php` as well as the `ContentFinder.php` file to prevent doubling of unnecessary code. E.g. I added a method which constructs a list of elements for the SQL `IN` keyword in the form of `(id1, id2, id3)` or `('string1', 'string2', 'string3')`. The code was already there but I think it is useful to encapsulate that functionality within a `protected` method and be available within the class hierarchy.

Additionally I updated all `quote(...)` and `quoteName(...)` calls to omit accessing the `$db` connection directly which makes the code more readable and smaller. It's a facade and as the `quote` methods are quite frequently used this makes sense to me to create two separate methods within the `Finder.php` to skip accessing the `$db` connection directly. Everything looks now much sleeker. I think this is everything.

Finally, I will now check within the Joomla code and the usage of tags in general but I have no doubts that this is the way to go. But I will report back when I checked that. Additionally I will update the Gantry 5 documentation as well to reflect the changes and new functionality.

mahagr:
Let me know when you've pushed the updated code; your changes are not yet visible as I'm guessing you wanted to check it out before.

thexmanxyz:
@mahagr I know, I had to wait on @Kubik-Rubik contribution invite. The changes are not pushed to the PR. You can now verify the changes. Everything is tested.

thexmanxyz:
@Kubik-Rubik @mahagr I refined the `.tags(...)` method to be consistent with the already existent finder API. 

I'm now using the plurals `ids` and `titles` for the parameters. Additionally I fixed a small bug in the original PR which resulted in a 2D array because in the initial commit you had to specify an array of `ids` in the form `tags({ids: [[id1, id2, id3]]})` which overcomplicates things.

Finally I updated the API so that we can also specify single values instead of arrays only:
`tags({ids: [ids]|id, titles: [titles]|title}, matchAll = false)`

To support the conversion I created a method which takes care of single values / collections and converts them into an array. The routine was already used by the `category(...)` method. I think we can leave it now how it is. I also updated my previous post which contains the full explanation to reflect these changes as well.

thexmanxyz:
@mahagr Here are the Joomla reference for using tags (with joins). This is done the same way like we are doing this:

- join `#__content` with `#__contentitem_tag_map`  for getting tag ids [here](https://github.com/joomla/joomla-cms/blob/659fa92727e0124836982ca773e1965613246e2a/administrator/components/com_content/models/featured.php#L257-L266)
- join `#__contentitem_tag_map` with `#__tags` for getting tag titles [here](https://github.com/joomla/joomla-cms/blob/fe597ef10fe6d6f0841bebc8aeebaad72d7080f3/modules/mod_tags_similar/helper.php#L80-L82)

I hope this is enough for verification. But what I noticed and what is missing currently within Gantry is `$db->quoteName(...)` for table names. Which is always present within Joomla. I will attach this to this PR at the positions were it is missing. I assume this is just for compatibility reason with some database system.

jameswadsworth:
This is so needed in Joomla! I hope after nearly 5 years this could be merge.

PR #2575, opened 2019-12-05
Menuitem link attributes rework, fixed #2573

### LINKED ISSUE #2573: Can't add rel="nofollow" to menu links
In the Gantry 5 menu editor there is no possibility to add rel="nofollow" to the link. Data attributes are applied at the LI level (which is no good).

original forum topic: https://rockettheme.com/forum/joomla-template-ambrosia/282803-nofollow-menu-link

thexmanxyz:
Hey, as brought up by @marktaylor46 (#2573) there is currently an issue with the `rel` attribute for the Gantry menu items in Joomla. The within the Joomla menu editor defined `rel` tag is not passed over to the Gantry menu. This seems to be a bug or was simply forgotten during the platform attribute implementation.

1.) Hence I fixed the **Menu.php** file for the Joomla platform and added the `rel` attribute which is now passed over to the **menu.html.twig** and **menuitem.yaml** file. I also added a dummy `rel` within the Grav platform **Menu.php** file, like it is currently handled for other attributes in that file.

2.) I added a `disabled` field in the **menuitem.yaml** file which displays the current `rel` attribute string passed over from the platform. Please see also my screenshot added on the bottom.

3.) If the link target for a menu item is `_blank` the `rel="noreferrer noopener"` tag will be automatically added. This functionality also cares if there is already a platform passed `rel` attribute and merges them correctly together without duplication. Also manually defined link attributes are considered. So if I want to extend the `rel` attribute with a custom string it is also merged into the `rel` attribute. See point 4.). 

4.) Additionally I adapted the functionality to empower users to add custom attributes on the `<a>` tag as it is already implemented for the `<li>` tag for menu items. 

5.) `<li>` as well as `<a>` attributes now contain the correct `exclude` definition so that now already existing attributes are not duplicated on this two tag types for menu items.

For further information on this PR please review the attached code. But I think this is everything that was changed.

grafik

### DIFF
--- a/engines/common/nucleus/particles/menu.html.twig
+++ b/engines/common/nucleus/particles/menu.html.twig
@@ -42,7 +42,13 @@
         {% set dropdown = item.level == 1 ? ' g-' ~ item.getDropdown() %}
         {% set parent = item.children ? ' g-parent' %}
         {% set target = (item.target != '_self' or context.particle.forceTarget) ? ' target="' ~ item.target|e ~ '"' %}
-        {% set rel = item.rel ? ' rel="' ~ item.rel|e ~ '"' %}
+        {% if item.target == '_blank' %}
+            {% set relTarget = ('noopener' not in item.rel) ? 'noopener' : '' %}
+            {% set relTarget = (relTarget) ? relTarget ~ ' ' : relTarget %}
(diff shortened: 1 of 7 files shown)


mahagr:
@thexmanxyz Merged with changelog updates

thexmanxyz:
@mahagr Thanks for merging :-)

PR #2641, opened 2020-05-07
fixed backend field alignment (line wrap)

thexmanxyz:
Hey, I suggest changing this one line in the admin CSS file to get correct alignment of the fields. IDK if this affects something else but it seems that the particle modal itself and the particle defaults tab should be all.

Before:
grafik

After:
grafik

Additionally this one also fixes this ugly "mobile" line-wrap.

Before:
grafik

After:
grafik

This issue happens because the `margin: 5px;` is not respected within the `max-width`. Also we create a `left-margin` for `.settings-param-field` of  `175px` which occupies the space but actually we would need `185px`. You can verify this in developers tools quite easily. I reduced the `max-width` to `165px` so we consume as much space as we have (`165px` + `10px` occupied by the margin, this is 2 x 5px).

### DIFF
--- a/platforms/common/scss/admin/_settings.scss
+++ b/platforms/common/scss/admin/_settings.scss
@@ -153,6 +153,7 @@
 
     .settings-param-title {
         max-width: 175px;
+        width: 175px;
         margin: 5px;
 
         .particle-label-subtype {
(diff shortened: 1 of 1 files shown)


monikat12titans:
I am closing this pull request because I could not find any backend field alignment issue, and there has been no activity since 2020.

PR #2665, opened 2020-06-21
Feature/theme config

jmgroud:
Hello
I want to create a new theme, for example from "Helium" to "NewTheme". By reading the documentation on this link ...
[http://docs.gantry.org/gantry5/theme-development/change-theme-name]
... I ask myself the following question: is it really necessary to modify absolutely ALL the character strings "Helium" by "NewTheme" in all text files? For example in the file "templateDetails.xml" is it not dangerous to modify the lines:

`<server type = "extension" priority = "1" name = "Helium Template"> http://updates.gantry.org/5.0/joomla/tpl_g5_helium.xml </server>`

by

`<server type = "extension" priority = "1" name = "Helium Template"> http://updates.gantry.org/5.0/joomla/tpl_g5_newtheme.xml </server>`

Are there other specific places where you have to be careful not to change these values?
If there is danger, perhaps it would be interesting to indicate it in the chapter "STEP 3" of the documentation ...


mahagr:
Not sure why you ended up doing a pull request against a branch which isn't finished.

You're right, that line should be removed instead of updated, but it's not really dangerous to leave it be as the call will most likely fail.

monikat12titans:
I am closing this pull request because there has been no activity since 2020.

PR #2723, opened 2020-10-23
[Wordpress] Load functions.php from custom folder

Chrissi2812:
This allows for fast and persistent overrides/additions to the template functions.php
This way the template can be updated without fear of loosing your functions.php changes.

Not sure how it works with child templates tho, as I never used them  (I'm a Wordpress noob 😅), but I know the custom folder acts as a kind of override system. 
My Usecase was that we needed to change the Gutenberg width, and the only way of doing it seems to add code to your functions.php file.

### DIFF
--- a/themes/helium/wordpress/functions.php
+++ b/themes/helium/wordpress/functions.php
@@ -54,3 +54,8 @@ foreach ($helpers as $file) {
 
     require $filepath;
 }
+
+// Custom Functions
+if ($customInclude = locate_template('custom/functions.php')) {
+    require $customInclude;
+}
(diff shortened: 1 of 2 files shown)


mahagr:
Thanks :)

mahagr:
@Chrissi2812 I did some improvements to allow you to alter the `$helpers` list from the overridden file. Can you test out the CI version and report back if it's good?

Chrissi2812:
I have tested your changes, and it worked just fine. Hadn't a chance to install the CI build at the moment, so have just changed the `functions.php` from the CI Build. Hope that helps.

Like that we can now just add `$helpers[] = 'custom/helpers/custom_post_types.php';` to include custom post types without an extra plugin :)

mahagr:
Yup, I thought it might be a good idea to change the order so you could do that. :)

Issue #2820, opened 2021-06-04
Remving Title Links from Joomla Article Particle


dmleeman:
Joomla 3.9.27
PHP 7.4.16
Gantry 5.4.37

I'm having some problem with Joomla and Gantry5 Joomla Articles particle. I have set up Modules in which the Joomla Article particle sits. Each Article has its own title and this is shown as the title of the article in the module/particle combo.
The problem is that I cannot remove the link from the article title, even through use of Joomla globals. So, all articles try to open on the home page as itemID101 from the link inherent in the title.
Has anyone come across this?
I am only using the default Joomla Articles particle that comes with base Gantry5

![image](https://user-images.githubusercontent.com/40118574/120842244-5a75c080-c564-11eb-9a90-0525abcc4248.png)

mahagr:
Particles generally use their own settings as they are independent of Joomla components and modules. After a quick look, I don't think there's an option to do this in the particle; maybe we need to add one?

dmleeman:
Given that Joomla itself allows the choice of linked titles then I would be a big fan of any change to allow the choice in such a fundamental Particle. The current method can seriously mess with SEO as Google bots try to follow the links. It's an avoidable situation.
Even an override would do for now.

mahagr:
Be free to make a pull request about this. :)

dmleeman:
Ah, yes :) I'm in the wrong place again. Thank you Matias

mahagr:
Seriously, if you find a way to improve Gantry, all contributions are welcome.

In this case, the easiest way is just to add an option into the blueprint to support this feature.

dmleeman:
Thanks Matias, will do. Still getting used to github and it's structure for such requests. So will deal with my In Tray and look at this later today.

dmleeman:
I think it is contentarray.html.twig in /media/gantry5/engines/nucleus/particles

{% if display.title.enabled %}
                                    <div class="g-array-item-title">
                                        <h3 class="g-item-title">
                                            <a href="{{ article.route }}">
                                                {{ display.title.limit ? article.title|truncate_text(display.title.limit) : article.title }}
                                            </a>
                                        </h3>
                                    </div>
                                {% endif %}
So, it is a case of removing the a href link on line 72 and 74

What do you think is the safest syntactically correct method? Simple comment?

mahagr:
I would just add an option for title link and display either `<a>` or `<span>` depending on the new setting.

dmleeman:
[...]

PR #2823, opened 2021-06-18
Removing Title Links from Joomla Article Particle #2820

dmleeman:
Joomla 3.9.27
PHP 7.4.16
Gantry 5.4.37

I'm having some problem with Joomla and Gantry5 Joomla Articles particle. I have set up Modules in which the Joomla Article particle sits. Each Article has its own title and this is shown as the title of the article in the module/particle combo.
The problem is that I cannot remove the link from the article title, even through use of Joomla globals. So, all articles try to open on the home page as itemID101 from the link inherent in the title.
Has anyone come across this?
I am only using the default Joomla Articles particle that comes with base Gantry5

image


mahagr:
There's something wrong with your pull request. Can you please change the target branch to `feature/v5.5.0` instead of `develop`?

mahagr:
Actually.. I don't think this has your changes at all.

dmleeman:
Yup, I'm in the wrong branch! Will figure this out carefully so I don't mess up again.

mahagr:
Just create a new pull request with the correct branches. I'm closing this one.

Issue #2913, opened 2021-10-19
FontAwesome4 icons do not work in Joomla 4


JoomFX:
The FA asset loaded by Joomla4 overrides the one loaded by Gantry5.
And since the FA asset that comes from Joomla4 always loads FA5, the icons get broken if the user has selected to use FA4 in the Gantry settings.

![j4-fa4](https://user-images.githubusercontent.com/7753940/137908877-a70d16d7-cdf7-4ec3-9e34-77e2153d9a1c.png)

@mahagr  I'll open a PR for this one as well :)

PR #2914, opened 2021-10-19
fix: wrong load order of fontawesome assets

JoomFX:
fixes https://github.com/gantry/gantry5/issues/2913

@mahagr modify it if `{% block head_overrides -%}` is not the best place :)


mahagr:
Nah, it's ok as it is

Issue #2979, opened 2022-01-07
bug in nucleus.css line 140 (img selector)


tetsuotram:
joomla 4.0.5  -  gantry 5

at line 140 in my nucleus.css (media/gantry5/engines/nucleus/css-compiled/nucleus.css), inside the img selector, there is "width: auto\9;" 
the part "\9" make the property unreadable from browser.

mahagr:
I removed all CSS hacks for IE 6-9.

PR #2981, opened 2022-01-10
Update nucleus.css

tetsuotram:
this solve the issue #2979

### DIFF
--- a/engines/common/nucleus/css-compiled/nucleus.css
+++ b/engines/common/nucleus/css-compiled/nucleus.css
@@ -137,7 +137,7 @@ sub {
 }
 
 img {
-  width: auto\9;
+  width: auto;
   height: auto;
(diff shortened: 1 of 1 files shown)


marktaylor46:
That's not actually a bug - it's a CSS hack for IE9. All other browser will ignore it.

tetsuotram:
thank you Mark for the reply!
But I can see that my Firefox is throwing an error: do you mean that I can safely ignore it?

marktaylor46:
Is it really an error or a warning? It's quite common for browser to issue warnings about things they don't understand - a warning you can ignore.

tetsuotram:
yes it is a warning.
thank you: I was not aware of this kind of hack

mahagr:
The line should really be removed as we have already dropped IE9 support.

PR #3133, opened 2023-04-01
Update composer.json

osnodegeoffrey:
To avoid ctype_digit deprecation


hexplor:
Already fixed but thanks!

PR #3192, opened 2023-10-20
Fix #3191

### LINKED ISSUE #3191: Particles in their own subdir show a non-descriptive error on Layout save
If you try to put particles into their own subfolder e.g. `particles/my-particle`, Gantry will crash at `preg_match()` with `Unknown modifier: m` on layout save.

Steps to reproduce:

1. Follow the steps in [the docs](https://docs.gantry.org/gantry5/advanced/creating-a-new-particle).
2. Create a subdirectory, e.g. `/particles/my-particle` and put your particle files there.
3. Use the particle in the Layout editor and save the layout.

Expected behavior:

It should save the layout without throwing an exception.

Possible fixes:
* Escape the path to the particle so that `preg_match` won't throw.
* Mention in the docs that particle files should be in a flat directory and either make the error message more descriptive or do not show the particles inside subdirectories in the Layout editor.

fcFn:
Proposed fix to #3191 to allow storing particles in their own directories for better organization.

### DIFF
--- a/src/classes/Gantry/Component/Layout/Version/Format2.php
+++ b/src/classes/Gantry/Component/Layout/Version/Format2.php
@@ -357,8 +357,8 @@ class Format2
             }
  
             // Check if type and subtype can be generated from the id.
-            if ($subtype && (preg_match("/".preg_quote( "/^{$type}-{$subtype}(-|$)/", "/" )."/", $id))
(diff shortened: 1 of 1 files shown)


fcFn:
While this does make the layout save and work correctly when using particles located inside a separate folder, options for such particles won't display. Not sure if this sort of enhancement is better than just making the error message more descriptive and/or adding to the docs not to create any subdirectories in the particles directory.

PR #3217, opened 2023-12-11
Fixed Changelog format

pmoreno-rodriguez:
Added extra spaces under the date and version 5.5.18


pmoreno-rodriguez:
Why the check fails?

pmoreno-rodriguez:
Apparently there is some conflict with this pull request

